Issue #208, opened 2014-12-30
Debian build process missing systemd file


thardie:
End of dpkg-buildpackage:

mv /home/users/thardie/exabgp/debian/exabgp/usr/lib/systemd /home/users/thardie/exabgp/debian/exabgp/lib
mv: cannot stat `/home/users/thardie/exabgp/debian/exabgp/usr/lib/systemd': No such file or directory
make[1]: *** [override_dh_auto_install] Error 1
make[1]: Leaving directory`/home/users/thardie/exabgp'
make: **\* [binary] Error 2
dpkg-buildpackage: error: fakeroot debian/rules binary gave error exit status 2

PR #209, opened 2014-12-30
debian: fix installation of systemd unit

vincentbernat:
I don't remember if we make Debian changes here or if Henry-Nicolas syncs its packaging from time to time.

Closes: #208


vincentbernat:
Unfortunately, this is too late. We can still push to backports once Jessie is released.

PR #210, opened 2015-01-02
Minor pull to fix bug with cli

lochiiconnectivity:
As per subject, also there is a typo on the qa code which is quite annoying


thomas-mangin:
@lochiiconnectivity - thank you David, the code is not even half complete tho :smile:

PR #215, opened 2015-01-15
Allow generic attribute not only in single line but also multiple lines

a16:
eg.)
```
neighbor 127.0.0.1 {
  --snip--
  static {
    route 111.111.111.111 {
      next-hop: 192.168.1.100;
      attribute: [ 0x99 0x70 0x00000064 ];
    }
  }
}
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,8 @@ Version explained:
 
 Version 3.4.7
  * Change: be more pythonic and use enty points with pip installation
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Thank you, I will merge but as it was not based on the latest code I can not do it via the web interface so it will have when I finished what I am working on.

a16:
It's my mistake. I forgot to follow the latest, Sorry. Thanks in advance.

PR #216, opened 2015-01-16
Fix healthcheck.py script to accept the loopback IP

charlescng:
IPs that are considered loopback IPs were not added to the addresses list. This caused the `healthcheck.py` script to not run.

@thomas-mangin

### DIFF
--- a/etc/exabgp/processes/healthcheck.py
+++ b/etc/exabgp/processes/healthcheck.py
@@ -236,7 +236,7 @@ def loopback_ips(label):
                 lmo = labelre.match(line)
                 if not lmo or not lmo.group("label").startswith(label):
                     continue
-            addresses.append(ip)
+        addresses.append(ip)
     if not addresses:
         raise RuntimeError("No loopback IP found")
     logger.debug("Loopback addresses: {0}".format(addresses))


thomas-mangin:
Thank you very much for this patch, I will wait for @vincentbernat to comment before pulling.

vincentbernat:
Well, I don't understand the problem nor the fix. `is_loopback` is true for `127.0.0.0/8`-like addresses. We don't want to advertise those. You either need to add some non-loopback addresses on `lo` or use `--ip` to tell which IP you want to advertise.

charlescng:
It was failing during the loopback IP retrieval.
```
processes/healthcheck.py --config healthcheck.conf --ip 192.168.99.2
DEBUG[healthcheck] Retrieve loopback IP addresses
ERROR[healthcheck] Uncatched exception: No loopback IP found
  File "processes/healthcheck.py", line 428, in <module>
    setup_ips(options.ips, options.label)
  File "processes/healthcheck.py", line 248, in setup_ips
    existing = set(loopback_ips(label))
  File "processes/healthcheck.py", line 242, in loopback_ips
    raise RuntimeError("No loopback IP found")
RuntimeError: No loopback IP found
```
I did not run into this when adding the `--no-ip-setup` flag to my setup, which is what I want to do regardless.

I don't think I quite follow the method's intentions but close this PR if you're happy with the code as-is.

Thanks for looking into it.

vincentbernat:
OK, I understand the problem better. I have opened another PR with the appropriate fix.

PR #217, opened 2015-01-19
healthcheck: don't raise an error when specifying IP on the command-line

vincentbernat:
When no loopback IPs are present and the IPs are specified on the
command-line instead, we still trigger an error due to a check in the
function retrieving the loopback IPs. We move this check in the outer
program instead. The check will fail only if there is no loopback IPs
and no IP specified on the command-line.

This should fix #216.


thomas-mangin:
Thank you @vincentbernat

PR #222, opened 2015-02-09
add as-path option

yeled:
allows one to prepend the as-path via an option. `--as-path '65044.24 65044.24 65044.24'` works for me.

caveat: You will break the upstream BGP peer if you try to prepend with an ASN that isn't your `local-as;`

ping @vincentbernat to check over this.


thomas-mangin:
Please Ignore the failing travis-ci .. it does not like my test ATM - not sure why.
I will let @vincentbernat comment before pulling but seems simple enough.

vincentbernat:
+1

PR #224, opened 2015-02-18
Fix for healthcheck.py for non-Linux systems

he32:
Two fixes for the non-Linux ifconfig route:
 * Also match "inet alias <address> netmask <netmask>", i.e. optionally match "alias ".
 * Don't insist on leading spaces, those are removed by strip() already...


vincentbernat:
LGTM.

thomas-mangin:
@he32 Thank you for this patch.

thomas-mangin:
@he32 hi, as I do not have your name could you please update the CHANGELOG to add attribution of your work :+1:

he32:
You're welcome.  I created pull request #228 for that.

PR #225, opened 2015-02-19
Update ancient.py to add send-packets flag to process

davidjwbbc:
During testing of an update to ExaBGP it was noticed that we were no longer getting the send-packets messages. This appears to be an omission in the flags transferred to the process object from the parsed configuration in lib/exabgp/configuration/ancient.py


thomas-mangin:
@davidjwbbc, Thank you David for fixing this omission - much appreciated.

PR #227, opened 2015-02-21
Removing forcing Attribute.CODE.ATOMIC_AGGREGATE into _generate_json loop.

icb-:
`_generate_json` in `exabgp.bgp.message.update.attribute.attributes.Attributes` loops through the keys of the dict given, *plus* `Attribute.CODE.ATOMIC_AGGREGATE`. Forcing that into the loop causes a `KeyError` if it wasn't already there. If it *was* already there, then adding it again isn't needed.

This also brings `_generate_json` in line with what `_generate_text` loops through.


thomas-mangin:
Thank you, not sure why I did that !

PR #228, opened 2015-02-23
Provide a pair of CHANGELOG entries for the healtheck.py fixes.

he32:


thomas-mangin:
Could not automatically merge - so done it by hand.

thomas-mangin:
Thank you for your help with ExaBGP

PR #229, opened 2015-02-24
bin/exabgp: simplify starter script

vincentbernat:
- Use of `readlink -f` to get absolute pathname to parent directory.
 - Use a for loop to test for available interpreters.
 - Unroll the Python script.
 - Select the first application matching directly in Python.
 - Quote everything appropriately in case someone put a space in a directory.

Tell me what you think of those changes and I can make the same in similar files.

### DIFF
--- a/sbin/exabgp
+++ b/sbin/exabgp
@@ -1,51 +1,22 @@
 #!/bin/sh
 
-dirname=`dirname $0`
+dirname="$(dirname $0)"
 
-case $dirname in
-	/*)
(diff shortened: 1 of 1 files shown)


thomas-mangin:
oh ... it is pretty :-) Thanks

vincentbernat:
I'll do the other scripts in the same way then.

thomas-mangin:
Thank you .. much appreciated.

PR #230, opened 2015-02-25
bin/*: simplify remaining starter scripts

vincentbernat:
Like for 41672c0cb56f, simplify other starter scripts:

 - Use of `readlink -f` to get absolute pathname to parent directory.
 - Use a for loop to test for available interpreters.
 - Remove support for python2.4 and python2.5 when they were still present.
 - Unroll the Python script.
 - Select the first application matching directly in Python.
 - Quote everything appropriately in case someone put a space in a directory.

### DIFF
--- a/bin/exabgp
+++ b/bin/exabgp
@@ -1,51 +1,20 @@
 #!/bin/sh
 
-dirname=`dirname $0`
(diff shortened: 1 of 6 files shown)


thomas-mangin:
Thank you

PR #236, opened 2015-03-07
Debian packaging update

vincentbernat:
Hey!

Updates for Debian packaging.

The important change is the fourth one, about not relying on the content of /etc/default/exabgp to start the daemon. This is considered bad practice. Instead, I rely on the presence of /etc/exabgp/exabgp.conf to start the daemon (including for systemd). The package doesn't ship `/etc/default/exabgp` anymore (useless) and we don't ship `/etc/exabgp/exabgp.conf` either. Previously, an empty configuration file was shipped. The post/preinst scripts are taking care of the migration. A user that did not modify `/etc/exabgp/exabgp.conf` will see the file removed while a user who did modify the configuration will keep it.


thomas-mangin:
@vincentbernat you should be able to push to the repository.
The dependencies presents in requirements.txt may be a good thing to add (as otherwise your code and the test suite will not run) respectively ipaddr and psutil
I think nose could be omitted.

thomas-mangin:
And I forgot to be polite and say thank you - I blame multitasking for it :tongue:

thomas-mangin:
I see that debian/exabgp.preinst does reference the latest version number, should this file be generated from setup.py when I tag a new release ? Or should the code be able to determine the version from the packaged files ?

vincentbernat:
No, it should stay as is. This is the first version where an empty `/etc/exabgp/exabgp.conf` is not shipped.

vincentbernat:
Nope, I cannot push directly to the repository. :)

thomas-mangin:
Should work now or I need a GitHub course :smile:

vincentbernat:
You're safe. It worked!

PR #239, opened 2015-03-08
debian: use manpages in `dev/` directory

vincentbernat:
I am doing a PR since I am unsure why the manpages are in `dev/`. Can they be used or is there something still missing? If they are ready, merge!


thomas-mangin:
Thank you, the reason the pages are in dev was that I was not sure where to put them myself.. dev does not seems the right place. As this is not debian specific, we could have a man/conf folder. I am open to suggestions.

vincentbernat:
Well, `dev/` is fine then. I am merging.

thomas-mangin:
I may change the location before release.

PR #241, opened 2015-03-14
tests: don't use assert, use self.assertXXX

vincentbernat:
In `qa/tests/control_test.py`, `assert` was still used. On failure, it's
not possible to know how the expected and the obtained results differ.


vincentbernat:
Oh, why?

thomas-mangin:
Because I am tired and I should stay away from computers :tired_face:

PR #242, opened 2015-03-21
Fix parsing attributes with PARTIAL flag set

dneiter:
Previously we were masking PARTIAL bit in attribute id instead of flag, which is wrong. As result, we were silently discarding received TRANSITIVE OPTIONAL attributes with PARTIAL flag set.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,10 @@ Version explained:
  minor : increase on risk of code breakage during a major release
  bug   : increase on bug or incremental changes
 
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Gosh ! very well spotted !

Can you please add an entry to the changelog with the attribution for the fix and I will pull.

dneiter:
Done. Added changelog entry.

thomas-mangin:
Thank you !

PR #249, opened 2015-05-01
add NSS_SDB_USE_CACHE to healthcheck.py command

mikemackintosh:
We saw an issue with nss-softtkn 3.14 and 3.16 where the excessive curling for the healthcheck command would cause available memory to be eaten up by dentry. 

Setting `NSS_SDB_USE_CACHE` in the `Popen` for the command portion ensured that curl would utilize the nss cache, and not continue to issue frivolous directory operations and eat up memory. 

See the following bug report for related information: https://bugzilla.redhat.com/show_bug.cgi?id=1044666

### DIFF
--- a/etc/exabgp/processes/healthcheck.py
+++ b/etc/exabgp/processes/healthcheck.py
@@ -284,6 +284,7 @@ def check (cmd, timeout):
 
     logger.debug("Checking command {0}".format(repr(cmd)))
     p = subprocess.Popen(cmd, shell=True,
+                         env = {"NSS_SDB_USE_CACHE":"YES"},
                          stdout=subprocess.PIPE,
                          stderr=subprocess.STDOUT,
                          preexec_fn=setpgrp_preexec_fn)


vincentbernat:
The bug report in Bugzilla is marked as solved, is the workaround still needed when a server is up-to-date?

Also, fixing in the script seems to be the wrong place. I mean, the bug is in nss-softkn. healthcheck.py is quite remote to get a fix about that. Moreover, the curl command is not directly used but provided by a user. You could run curl through `env NSS_SDB_USE_CACHE=yes curl ...` or you could run healthcheck.py with `env NSS_SDB_USE_CACHE=yes python healthcheck.py ...` or you could run exabgp with those environment variable set.

I may be a bit confusing. My point is that setting the environment variable for one specific use-case seems the wrong way to fix the problem. It could also create additional problems for users wanting to run commands where this environment variable must not be set.

thomas-mangin:
Hello Mike, Thank you for taking the time to open this bug and proposed a patch. It is much appreciated.

As stated by Vincent, this issue reported is not related to ExaBGP, or the health script, and I believe we should not include workaround for a third party application into our code and it indeed looks like that the issue could be solved better by wrapping the command within a small shell script such as :

#!/bin/sh
NSS_SDB_USE_CACHE=YES exec < your previous command >

We could however document this - perhaps adding some wiki page for healthcheck ... ExaBGP documentation is in dire need of love ..

Until this is done, I will not close this thread to remind me/Vincent/any good soul ..

PR #251, opened 2015-05-07
batch of changes related to bagpipe

tmmorin:
some changes are not required for ExaBGP but only for BaGPipe BGP
some changes are also required for ExaBGP to properly work (e.g. fixes in RTC class)
some others changes are cosmetic

(please tell me if you prefer multiple pull requests)

### DIFF
--- a/lib/exabgp/bgp/message/update/attribute/aggregator.py
+++ b/lib/exabgp/bgp/message/update/attribute/aggregator.py
@@ -36,7 +36,7 @@ class Aggregator (Attribute):
 	def __len__ (self):
 		raise RuntimeError('size can be 6 or 8 - we can not say - or can we ?')
 
-	def __str__ (self):
+	def __repr__ (self):
 		if not self._str:
 			self._str = '%s:%s' % (self.asn,self.speaker)
 		return self._str
--- a/lib/exabgp/bgp/message/update/attribute/aigp.py
+++ b/lib/exabgp/bgp/message/update/attribute/aigp.py
@@ -77,7 +77,7 @@ class AIGP (Attribute):
 			return self.packed
 		return ''
 
-	def __str__ (self):
+	def __repr__ (self):
 		return '0x' + ''.join('%02x' % ord(_) for _ in self.aigp[-8:])
 
 	@classmethod
--- a/lib/exabgp/bgp/message/update/attribute/aspath.py
+++ b/lib/exabgp/bgp/message/update/attribute/aspath.py
@@ -94,7 +94,7 @@ class ASPath (Attribute):
 	def __len__ (self):
 		raise RuntimeError('it makes no sense to ask for the size of this object')
 
-	def __str__ (self, confed=False):
+	def __repr__ (self, confed=False):
 		if self._str:
 			return self._str
 
--- a/lib/exabgp/bgp/message/update/attribute/atomicaggregate.py
+++ b/lib/exabgp/bgp/message/update/attribute/atomicaggregate.py
@@ -26,7 +26,7 @@ class AtomicAggregate (Attribute):
 	def __len__ (self):
 		return 0
 
-	def __str__ (self):
+	def __repr__ (self):
 		return ''
 
 	def __cmp__ (self, other):
--- a/lib/exabgp/bgp/message/update/attribute/attribute.py
+++ b/lib/exabgp/bgp/message/update/attribute/attribute.py
@@ -103,11 +103,8 @@ class Attribute (object):
 			0xffff:             'internal-split',
 		}
 
-		def __str__ (self):
-			return self.names.get(self,'unknown-attribute-%s' % hex(self))
-
 		def __repr__ (self):
-			return str(self)
+			return self.names.get(self,'unknown-attribute-%s' % hex(self))
 
 		@classmethod
 		def name (cls, self):
--- a/lib/exabgp/bgp/message/update/attribute/attributes.py
+++ b/lib/exabgp/bgp/message/update/attribute/attributes.py
@@ -216,7 +216,7 @@ class Attributes (dict):
 			self._json = ', '.join(self._generate_json())
 		return self._json
 
-	def __str__ (self):
+	def __repr__ (self):
 		if not self._str:
 			self._str = ''.join(self._generate_text())
 		return self._str
--- a/lib/exabgp/bgp/message/update/attribute/clusterlist.py
+++ b/lib/exabgp/bgp/message/update/attribute/clusterlist.py
@@ -37,7 +37,7 @@ class ClusterList (Attribute):
 	def __len__ (self):
 		return self._len
 
-	def __str__ (self):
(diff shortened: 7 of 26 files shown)


thomas-mangin:
No all in one is good - I will just need to review them. Thank you Thomas.

thomas-mangin:
You change a lot of __str__ to __repr__ in the code, this may be good for your case but it will definitively break ExaBGP .. Ideally, the code should NOT rely on __str__ for creation of string the the API but in practice currently it does so removing the __str__ will cause mayhem !

thomas-mangin:
could you instead add a new __repr__ which return str(self) ?

thomas-mangin:
Would you also be kind enough to explain to me why you are using repr instead of str in your own code ?

thomas-mangin:
Hello @tmmorin  - merged everything but the __repr__ changes.

I propose to change ExaBGP to:
 -  not use __str__ (find a new name)
 - cause every __str__ to raise a Deprecated Message while calling the new function.
 - update my code and your to the new API 

What do you think ? Using __str__ was a bad idea in the first place but it took ExaBGP to become a large(r) program for me to realise it. As I am working on 3.5, this is the ideal time for fix this issue.

thomas-mangin:
https://www.codacy.com/app/thomasmangin/exabgp_2/issues?&filters=W3siaWQiOiJDYXRlZ29yeSIsInZhbHVlcyI6WyJFcnJvciBQcm9uZSJdfV0=

The two last issues are in the new code

tmmorin:
Thanks for merging these pieces!

There are two reasons why I ended up changing ExaBGP behavior wrt. __repr__ and __str__:
* some `__hash__`, `__eq__`, `__ne__` or `__cmp__` implementations actually rely on repr or str to expose the right things to behave properly wrt. to how bagpipe uses nlri, attributes and RTs in sets and dicts
* bagpipe logs use repr on some of these objects

The reason why I proposed to use repr instead of str is the following: str(iterable) calls repr on the internal objects, thus str(multivalue attribute) ends up calling repr on each attribute (e.g. for extended communities).  By replacing all the `__str__` definitions by `__repr__` definitions, I get the wanted behavior, *and* `__str__` still works because if __str__ is undefined but __repr__ is Python calls __repr__ (Python does not do it the reverse way: if __repr__ is undefined, repr(object) displays the object address, which is useless).

I would still like `__repr__` or `__str__` to return something useful, if you end up deciding that __str__ should be used, this would be also fine for bagpipe, but for extended communities I would need a tweak on bagpipe side so that str is called on members of attributes[Attribute.CODE.EXTENDED_COMMUNITY].communities.

Another thing is, if you drop __str__ and __repr__, bagpipe would need that `__hash__`, `__eq__`, `__ne__` and `__cmp__` behavior is made consistent to allow storing objects in sets and/or using them in hashes.  We may need a discussion focused on this particular issue because, the above pull request still does not get the right behavior for bagpipe:  
* example 1: hash and cmp defined in MPLSVPN are good, but cmp is in fact not called when MPLSVPN objects are used in sets because eq/neq defined in MPLS are defined and used preferably. I can override eq/neq in MPLSVPN
* example 2: using RTs as dict keys or in sets would not work because cmp defined in extended.Community is not called, because eq/neq defined in Attribute is preferred ; I can override eq/neq in ExtendedCommunity or in RouteTarget, but I fear that the behavior would not be consistent with the behavior of eq of Attribute (which tests only attribute id, but not the value); my question is: do the Attribute and Attributes classes depend on this behavior of Attribute.eq or could we change this ?

thomas-mangin:
Are you saying that \_\_repr\_\_ should return a object which can be used by __cmp__ ?
If so we should do this and have an \_\_repr\_\_ object which is not the same as what is printed.

I have a wish to undefine all \_\_str\_\_ object and make the printing explicit. I believe using the built-in lead to some issue when it comes to provide backward compatibility as it can not be passed a version number ... 

Your opinion ? If I have to perform such intrusive change 3.5.0 - which is where we are - is going to be the right moment for it !

thomas-mangin:
> my question is: do the Attribute and Attributes classes depend on this behavior of Attribute.eq or could we change this ?

I am pretty sure I rely on __eq__ to check if two routes are equals. I would need to dig the code. I have only define comparison when it was required - many object have them as I ported your code as it and did not want to remove it.

tmmorin:
> Are you saying that __repr__ should return a object which can be used by __cmp__ ? If so we should do this and have an __repr__ object which is not the same as what is printed.

No, my idea would be:
- have a nice repr usable for logs and not implement str unless we want str to output something else than repr 
- cover all classes with hash/cmp implementations that allow proper testing of equality of objects and using them in sets or as dict keys  (implementing cmp alone allows to not implement eq and neq, Python will derive eq and neq from the result of cmp).
- multivalued attributes (such as ext coms) will also need to be comparable (see Attributes.sameValuesAs)

The last item I hope we can achieve, but we might fail if exabgp have different needs when comparing objects. The one typical case that I have in mind is the following: bagpipe stores nlris advertised by a peer in a set, a relies on set matching to identify that an already advertised route is withdrawn ; for this to work for MPLS labelled routes, eq and cmp have to *ignore* the label field. If this is fine for ExaBGP then we are all good, but if exabgp needs to have two MPLS NLRIs differing by their label to be inequal, then the solution will not be trivial.

> I have a wish to undefine all __str__ object and make the printing explicit. I believe using the built-in lead to some issue when it comes to provide backward compatibility as it can not be passed a version number ... 

I don't see such a need right now for bagpipe.
You know better for ExaBGP, of course.
Maybe you could have a new method, versioned for exabgp use, that would default to calling str or repr ?

> I am pretty sure I rely on __eq__ to check if two routes are equals. I would need to dig the code. I have only define comparison when it was required - many object have them as I ported your code as it and did not want to remove it.

Attribute.__eq_ only compare the attribute id, but does currently not look at the attribute value. 
If used to compare route equality, it will not give the expected result.

thomas-mangin:
Would this be acceptable ?
 (1) I first need to make sure ALL classes LOOSE the \_\_str\_\_, make it raise a RuntimeError
 (2) replace it with a def text(version)
 (3) then we can add the \_\_repr\_\_ you need which does not need to conform to the text api interface used by exabgp. 
 (4) Ideally any format which need to not change for bagpipe should be another function name so it does not rely on any python "magic".
 (5) perhaps remove all the \_\_str\_\_ functions ...

For the comparison, I need to look at how your use case but I am sure we can find a way which works for both implementation. I do little comparison of objects and it to know if two attributes are the same or if a route need to be withdrawn/announced so our constraint must be similar.

tmmorin:
For str/repr and text(),yes, that looks good.

And for the comparison, it's great if you are able to conclude that our needs are the same. I can push to you the changes I would have for eq/neq/cmp/hash and you will see if you think this will work well for ExaBGP.

thomas-mangin:
I am closing the pull request as nothing more can be pulled here.

PR #255, opened 2015-05-08
Use upstart along with SysV init

pyke369:
b/c we want exabgp to be supervised and restarted in case of crash


thomas-mangin:
Hello, thank you for taking the time to submit this pull request.
The debian packaging done by @vincentbernat - so I will wait for his comment before pulling but I note that AFAIK that Ubuntu moved to systemd with Vivid Vervet so I believe that more control of what is installed depending on the OS may be required.

pyke369:
Thanks for your note Thomas.

This PR was specifically crafted for older version of Ubuntu, where systemd - while being shipped - is not the default init system yet (I'm talking about Precise and Trusty LTS).

Using Upstart on those versions is a nice improvement compared to SYSV init, especially when ExaBGP is mission-critical to your datacenter/services platform SDN strategy (which is our case, and we unfortunately had a few crashes in the past).

(and I'm taking the opportunity to thank you for this amazing piece of software by the way \o/).

thomas-mangin:
Hi Pierre-Yves. I/we do run ExaBGP in production with Ubuntu under daemontools or supervisord ( we are currently migrating ). I understand the need to support the current park, including LTS. The issue is that the patch would break Debian which does use Upstart. So both scripts surely need to be present in the repository and depending on the machine generating the .deb include one or the other. As I am not an packaging expert ( I must have built my last SRPM 10 years ago ! ) I am waiting from @vincentbernat feedback to see how we could best move forward.

pyke369:
Understood, makes perfect sense. I guess we could leave the "init.d" file in the debian folder, along with "exabgp.upstart" (although this naming is quite new to me, I though Debian packaging guidelines dictated "<pkgname>.init" and "<pkgname>.upstart"...). Let's wait for Vincent's opinion on this.

vincentbernat:
Hey @pyke369 !

Yes, leave the SysV file. When both of them are present, Upstart will use the file in `/etc/init`.

As for naming, you can rename `init.d` to `exabgp.init`. Since there is only a single package, the package name is not mandatory but it will match the other files in the folder. `init.d` is a valid suffix for `dh_installinit` but it is not documented.

pyke369:
I made the appropriate changes to the pull request.

vincentbernat:
LGTM.

PR #273, opened 2015-05-20
various changes related to MPLSVPN and E-VPN

tmmorin:


lib/exabgp/bgp/message/update/attribute/mprnlri.py
```diff
@@ -171,6 +174,10 @@ def unpack (cls, data, negotiated):
 				length,nlri = NLRI.unpack(afi,safi,data,addpath,'',IN.ANNOUNCED)
 				nlris.append(nlri)
 
+			if length == 0:
```
thomas-mangin: Why the length of the MP is recaculated here please ?

lib/exabgp/bgp/message/update/nlri/evpn/mac.py
```diff
@@ -42,8 +42,12 @@ class MAC (EVPN):
 	NAME = "MAC/IP advertisement"
 	SHORT_NAME = "MACAdv"
 
-	def __init__ (self, rd, esi, etag, mac, maclen, label, ip, packed=None):
-		EVPN.__init__(self, packed)
+	def __init__ (self, rd, esi, etag, mac, maclen, label,ip,packed=None,nexthop=None,action=None,addpath=None):
+		EVPN.__init__(self,packed,nexthop,action,addpath)
+		assert(isinstance(rd, RouteDistinguisher))
```
thomas-mangin: I will comment those asserts

lib/exabgp/bgp/message/update/nlri/evpn/multicast.py
```diff
@@ -30,16 +29,20 @@ class Multicast (EVPN):
 	NAME = "Inclusive Multicast Ethernet Tag"
 	SHORT_NAME = "Multicast"
 
-	def __init__ (self, rd, etag, ip):
+	def __init__ (self, rd, etag, ip, packed=None,nexthop=None,action=None,addpath=None):
+		EVPN.__init__(self,packed,nexthop,action,addpath)
+		assert(isinstance(rd, RouteDistinguisher))
```
thomas-mangin: I will also comment those assert

lib/exabgp/bgp/message/update/nlri/qualifier/labels.py
```diff
@@ -50,15 +50,24 @@ def __str__ (self):
 			return ' label %s' % self.labels[0]
 		else:
 			return ''
+		
+	def __repr__(self):
+		if self._len > 1:
+			return '[%s]' % ','.join([str(_) for _ in self.labels])
```
thomas-mangin: needs spaces around %s, i.e. [ %s ]

lib/exabgp/bgp/message/update/attribute/mprnlri.py
tmmorin: I the the subclass on which we called unpack has, for some reason, returned a zero length, then unless we do soemthing like the above, we will call it again with the same data and it will return zero again... loop... :(

So let's consider instead that if len==0 then it means it has consumed everything...

lib/exabgp/bgp/message/update/nlri/evpn/mac.py
tmmorin: As you wish. I thought they were helpful safeguard and that the impact could be zero at runtime by starting exabgp with "python -0".

lib/exabgp/bgp/message/update/nlri/qualifier/labels.py
tmmorin: ok

tmmorin:
Thanks a lot for merging !

lib/exabgp/bgp/message/update/nlri/evpn/mac.py
thomas-mangin: when pip is installing the application I do not know how the code ends ups. The #! script starting python itself is not run with -0

lib/exabgp/bgp/message/update/attribute/mprnlri.py
thomas-mangin: If the subclass is buggy the code should be fixed in the buggy sub-class not the caller. Can I remove this code please ?

tmmorin: Ok.  But at least something like the following would be in order:

if length == 0: raise Exception("sub-calls should not return length zero")

You don't want a bogus BGP route triggering a parsing bug itself triggering an endless loop, so it's nice to have a safeguard.

thomas-mangin: Sure, I will update the code to raise a RuntimeError ..

PR #286, opened 2015-06-04
Fix encoding/decoding of transitive bit in extended communities

tmmorin:
There is a trick: bit set means "not transitive".

RFC4360:
        T - Transitive bit
            Value 0: The community is transitive across ASes
            Value 1: The community is non-transitive across ASes

(This was found, and the fix confirmed, by interop testing with Contrail BGP implementation.)

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/286)

### DIFF
--- a/lib/exabgp/bgp/message/update/attribute/community/extended/community.py
+++ b/lib/exabgp/bgp/message/update/attribute/community/extended/community.py
@@ -8,6 +8,8 @@ Copyright (c) 2009-2015 Exa Networks. All rights reserved.
 
 from exabgp.bgp.message.update.attribute import Attribute
 
+from struct import pack
+
 # ======================================================= ExtendedCommunity (16)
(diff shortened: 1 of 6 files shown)


thomas-mangin:
Hi Thomas,

well spotted ! I really need you test suite ! Could we have the code changed 
from :
self.COMMUNITY_TYPE | 0x40 if transitive else self.COMMUNITY_TYPE,0x02
to : 
self.COMMUNITY_TYPEif transitive else self.COMMUNITY_TYPE  | 0x40,0x02
and not :
self.COMMUNITY_TYPE | 0x40 if not transitive else self.COMMUNITY_TYPE,0x02

tmmorin:
Sure, I'll make the change and re-pull-request.

Note that this was found not by a testsuite, but by manually testing interop with another BGP stack.

PR #287, opened 2015-06-04
comment inconsistent router_id check

tmmorin:
We can't enforce this check: it will fail if we try to setup a connection between two ExaBGPs. The reason is that we generate our own Open messages with neighbor.router_id, which cannot be at the same time our id and our neighbor's. I think Neighbor needs an additional field to store our neighbor's router-id to enable this.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/287)

### DIFF
--- a/lib/exabgp/bgp/message/open/capability/negotiated.py
+++ b/lib/exabgp/bgp/message/open/capability/negotiated.py
@@ -125,10 +125,16 @@ class Negotiated (object):
 		if self.received_open.router_id == RouterID('0.0.0.0'):
 			return (2,3,'0.0.0.0 is an invalid router_id')
 
-		if self.received_open.asn == neighbor.local_as:
-			# router-id must be unique within an ASN
-			if self.received_open.router_id == neighbor.router_id:
-				return (2,3,'BGP Indendifier collision, same router-id (%s) on both side of this IBGP session' % self.received_open.router_id)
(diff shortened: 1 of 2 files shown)


thomas-mangin:
I do not understand what is wrong ... I have many ExaBGP connecting to each other in /qa/self/

This test enforce unique router-id within an IGP which is right ( router-id should be globally unique but there is discussion to only make it unique within an IGP - which is what we are doing ), no two configuration should use the same router-id.

lib/exabgp/bgp/message/open/capability/negotiated.py
```diff
@@ -125,10 +125,16 @@ def validate (self, neighbor):
 		if self.received_open.router_id == RouterID('0.0.0.0'):
 			return (2,3,'0.0.0.0 is an invalid router_id')
 
-		if self.received_open.asn == neighbor.local_as:
```
thomas-mangin: if self.received_open.asn == neighbor.local_as:

self.received_open is the decoded open our peer sent us, so it will have our peer's router-id and ASN
neighbor.local_as, is our ASN, so this check make sure it is an IBGP session.

if self.received_open.router_id == neighbor.router_id:

We are checking that our router-id is not the same their ( unique within the IGP ).

tmmorin:
What you say makes sense, I have to investigate why in my setup this test ended up failing, and why I got confused.

PR #288, opened 2015-06-05
Fixes encoding/decoding of transitive bit in extended communities

tmmorin:
There is a trick: bit set means "not transitive". ::

>>   RFC4360:
>>        T - Transitive bit
>>  -->         Value 0: The community is transitive across ASes
>>               Value 1: The community is non-transitive across ASes

This patch:
* factors out the code to packed community type and subtype
* avoids hardcoded 0x.. values in each ext com
* makes this code produce the right bit value for the transitive bit (applies only to RT at this point)
* fixes Origin ext com packing which was using wrong subtype (0x02 instead of 0x03)

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/288)

### DIFF
--- a/lib/exabgp/bgp/message/update/attribute/community/extended/community.py
+++ b/lib/exabgp/bgp/message/update/attribute/community/extended/community.py
@@ -8,6 +8,8 @@ Copyright (c) 2009-2015 Exa Networks. All rights reserved.
 
 from exabgp.bgp.message.update.attribute import Attribute
 
+from struct import pack
+
 # ======================================================= ExtendedCommunity (16)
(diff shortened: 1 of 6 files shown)


thomas-mangin:
Much better patch :-) Your naming convention are a little bit long and Java'ish :wink: please excuse me if I decide to rename your function something shorter at some point in time :angel:

tmmorin:
no problem :)

thomas-mangin:
Thomas, could you please backport this fix to the 3.4 branch ? if you do not have the time, I will do it .. under duress ..

tmmorin:
Just did !

PR #305, opened 2015-07-17
Update list of users

plajjan:
Add Deutsche Telekom / TeraSteram, which is where I work. We don't have
any public reference but we are using ExaBGP.

Fixed upper/lower case as my eyes were bleeding from all that SCREAMING ;)

Also sorted in alphabetical order which I think makes sense since I
don't think there is any inherent order to the entries.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/305)


thomas-mangin:
Thank you very much for this patch :+1: :smile:

PR #307, opened 2015-07-19
one more NLRI packing fix

tmmorin:
One more occurence of packed instead of _packed.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/307)


thomas-mangin:
I will merge, I have a MASSIVE NLRI patch pending on my laptop (but it should not conflict) so I will apply once I have finished (I have been working on it on and off for the last week)

PR #308, opened 2015-07-19
NLRI unit tests for MPLSVPN and EVPN

tmmorin:
[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/308)

### DIFF
--- a/lib/exabgp/bgp/message/update/nlri/evpn/mac.py
+++ b/lib/exabgp/bgp/message/update/nlri/evpn/mac.py
@@ -92,7 +92,7 @@ class MAC (EVPN):
 	def _pack (self):
 		if not self._packed:
 			ip = self.ip.pack() if self.ip else ''
-			self.packed = "%s%s%s%s%s%s%s%s" % (
+			self._packed = "%s%s%s%s%s%s%s%s" % (
 				self.rd.pack(),
 				self.esi.pack(),
 				self.etag.pack(),
--- a/qa/tests/nlri_tests.py
+++ b/qa/tests/nlri_tests.py
@@ -1,0 +1,317 @@
+#!/usr/bin/env python
+# encoding: utf-8
+"""
+open.py
+
+Created by Thomas Morin, Orange on 2015-07-10.
+Copyright (c) 2009-2015 Orange. All rights reserved.
+"""
+import unittest
+
+from exabgp.configuration.setup import environment
+environment.setup('')
+
+from exabgp.reactor.protocol import AFI, SAFI
+
+from exabgp.bgp.message.update import Attributes
+
+from exabgp.bgp.message.update.attribute.localpref import LocalPreference
+from exabgp.bgp.message.update.attribute.community.extended.communities \
+    import ExtendedCommunities
+from exabgp.bgp.message.update.attribute.community.extended \
+    import RouteTargetASN2Number as RouteTarget
+from exabgp.bgp.message.update.attribute.community.extended.encapsulation \
+    import Encapsulation
+
+from exabgp.bgp.message.update.nlri.mpls import MPLSVPN
+from exabgp.bgp.message.update.nlri.evpn.mac import MAC as EVPNMAC
+from exabgp.bgp.message.update.nlri.qualifier.rd import RouteDistinguisher
+from exabgp.bgp.message.update.nlri.qualifier.labels import Labels
+from exabgp.bgp.message.update.nlri.qualifier.esi import ESI
+from exabgp.bgp.message.update.nlri.qualifier.etag import EthernetTag
+from exabgp.bgp.message.update.nlri.qualifier.mac import MAC
+
+from exabgp.protocol.ip import IP
+
+from exabgp.bgp.message import OUT
+
+
+
+def prefixToPackedIPMask(prefix):
+    ipString, mask = prefix.split("/")
+    return (IP.pton(ipString), int(mask))
+
+
+class TestNLRIs(unittest.TestCase):
+
+    # tests on MPLS VPN NLRIs
+
+    def test0_MPLSVPNHashEqual(self):
+        '''
+        Two indistinct VPN NLRIs should
+        hash to the same value, and be equal
+        '''
+        rd = RouteDistinguisher.fromElements("42.42.42.42", 5)
+
+        packedPrefix, mask = prefixToPackedIPMask("1.1.1.1/32")
+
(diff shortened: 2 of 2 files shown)


thomas-mangin:
Same with this patch ... It will conflict and would like to discuss with you how you want to handle the new classes.

tmmorin:
Go ahead, tell me :)

thomas-mangin:
I need to have the patch in my tree to show you first and works keep getting in the way. The code is mostly there but I need to iron a few last bug to make sure all the unit tests are passing..

thomas-mangin:
I made some time to finish the code tonight - now I must check if it will have caused some wrath home ..
The code is not yet pushed to my tree until you can check.
https://github.com/thomas-mangin/exabgp/commit/ceb315ca380a0d0d40d3132fb05769c53cd3fd1c

thomas-mangin:
I have merged your patch into mine. As you can see there is a change in API ( an explicit .cidr ).
I would like to know if the changes done to the class MPLSVPN should be merged into the class MPLS (which ExaBGP use) or if there is a reason to have two classes.

tmmorin:
A few comments:
- a reason to have a different MPLSVPN class is for route creation ; I would find it confusing to use the same class to create routes of different AFI, SAFI based on the presence or not of an RD
- bagpipe-bgp needs a constructor for MPLSVPN that allows to specify the prefix, but also the RD and the label  -- this ability is lost in your refactoring
- the hash method for MPLSVPN has to be consistent with eq and thus can't use the label

By the way, related to the comment above, if I create a route with MPLSVPN, how does the code guess which AFI to use ? does it test the AFI of the prefix ?

tmmorin:
One precision: the MPLSVPN constructor would also need to allow specifying the nexthop and the action.

thomas-mangin:
Hi Thomas,

Thank you for your answers. The AFI/SAFI is always explicitly stored in NLRI objects.
 - from the parser, it is inferred but very soon ExaBGP will have one section per AFI/SAFI routes (as it should be explicit)
 - from the wire, the information is in the MP(U)RNLRI and then set

The "normal" ExaBGP's MPLS class has an RD. it is set to .NORD (i.e.: not present) by default and therefore the RD is not encoded if absent. There is no issue using the same class for both as the RD does not affect the encoding ( it is skipped if set to NORD - using "is" for comparaison ).

The new code with a simpler \_\_init\_\_ (  afi, safi, action ) makes the rest of the code much better but I would be happy creating a @classmethod for every NLRI class as an extra, stable, API for bagpipe-bgp. So, for example, we could have do MPLS.new( .... ) , INET.new ( .... )  with an API of your choice or matching what was changed - including nexthop.

I do not believe the hash is an issue and MPLS could use the hash you need.

thomas-mangin:
Let me put a patch, before looking at the code again ..

thomas-mangin:
Could it be I now called Label - what you call MPLS and MPLS what you call MPLSVPN ?

tmmorin:
Having stable .new classmethods to build objects from bagpipe is a good idea.

All this looks good to me.
Once the new constructors are there I will update the unittests.

One the names:
- the class common to labelled prefixes and labelled VPN prefixes, could be called "LabelledPrefix", but Label alone would not I think be a good name
- it would be misleading to name the class for VPN routes "MPLS", because non-VPN labelled routes are MPLS routes ; I think a good name would be LabelledVPNPrefix, another possibility is "IPVPN"
- if you prefer shorter names, we could drop Prefix and keep "Labelled" for the common class, and opt for "IPVPN" 

About the hash (and eq):
- we agree, I was just commenting your "XXX: Don't we need to have the label here ?" in mplsvpn.py 
- if we look beyond bagpipe, the same eq/hash would make sense for non-VPN labelled routes, for which comparisons/hashing should also *not*  take the label into account
- one thing we could do, *if* bagpipe would end up having different needs, is have bagpipe us NLRI.register to register subclasses with a different eq/hash behavior

thomas-mangin:
I will look at renaming the classes to have clearer names. I will have a think. My initial instinct would be:
 - ends in NLRI
 - prepend Label(led) when a label is present
 - prepend VPN when a route distinguisher is present

Also, I wrote this patch for bagpipe
https://github.com/Exa-Networks/exabgp/commit/31270ef79c0f0fcd1e23df26ef3f4f5f04bc3b8c
This will allow you to change which class are registered with the decoder for each NLRI.
```
> from exabgp.bgp.message.update.nlri import INET
> from exabgp.protocol.family import AFI
> from exabgp.protocol.family import SAFI
> @register(AFI.ipv4,SAFI.unicast,True)
> class MyINET (INET): pass
```

thomas-mangin:
So the class name could change to become :
 - `INET` --> `INET`
 - `Label` --> `Labelled` ( I am not sure about this one )
 - `MPLS` --> `IPVPN`
 - `MPLSVPN` --> `MPLS`

tmmorin:
Ok, I'm lost. In fact I hadn't noticed the Label class and thus wrongly thought that the MPLS class (which was really badly named) was factoring common code between labelled routes and labelled VPN routes.

Let me try to summarize my understanding:
- INET: ok
- Label: just has labels and cidr => ok, this is a labelled prefix (RFC3107 ), it would be ok to call it Labelled  (it is not bound to AFI.ipv6/SAFI.nlri_mpls, but I think it should, like for ipv4)
- MPLS: adds RDs to Label, so yes, it would be well named as IPVPN (the initial name was very wrong), *however* it should *not* be bound to SAFI.nlri_mpls (which has no RDs)
- MPLSVPN: seem to be useless assuming that MPLS renamed as IPVPN (my initial motivation for doing a new class was to be sure not to break your existing code for init eq hash, etc.)

I hope that you have a great weather during your holidays and that you will ignore this comment until you get back!  ;-)

thomas-mangin:
Thank you @tmmorin for the clarification - It helped. I will perform these changes .. if the weather gives me a reason to do so !

So to summarise : 
 - rename Label to Labelled, MPLS to IPVPN, remove MPLSVPN
 - add a new @classmethod for full object creation (will look at the 3.4 interface to keep it the same)
 - use bagpipe friendly hash functions
 - add a new function to compare two NLRI without taking in consideration action and nexthop .. oh ! no ! I need to find a new name for this too !

tmmorin:
I'm ok to not have the bagpipe-friendly specific eq/cmp/repr in exabgp, the new register decorator looks fine.

thomas-mangin:
@tmmorin 
- rename : done
- new class method (called new) : done pour IPVPN (do you need other classes ?)
- bagpipe friendly hashes : done
- compare (added eq) : done

PR #319, opened 2015-08-19
systemd: make exabgp runs as exabgp:exabgp with systemd

vincentbernat:
Both SysV init script and upstart job are using exabgp:exabgp as a user
to run exabgp. Do the same thing for systemd.

Made as a PR because the systemd script is not distribution-specific. We could also create the appropriate file to let systemd create this user/group if it doesn't exist.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/319)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -83,6 +83,8 @@ Version 4.0.0
    reported by: Dmitry Onuchin
  * Fix: reading large buffered data from helper process caused truncation
    reported by: qqTYXn7
+ * Change: ExaBGP is now run as user/group exabgp/exabgp with the systemd
+   service file.
 
 Version 3.4.10
  * Fix: Fix parsing attributes with PARTIAL flag set
--- a/etc/systemd/exabgp.service
+++ b/etc/systemd/exabgp.service
@@ -3,6 +3,8 @@ Description=ExaBGP
 After=network.target
 
 [Service]
+User=exabgp
+Group=exabgp
 Environment=exabgp_daemon_daemonize=false
 Environment=exabgp_log_destination=stdout
 Environment=ETC=/etc
--- a/etc/sysusers.d/exabgp.conf
+++ b/etc/sysusers.d/exabgp.conf
@@ -1,0 +1,1 @@
+u exabgp - "ExaBGP"


thomas-mangin:
Seems like a good idea ..
Can you please add an entry into the CHANGELOG file to indicate this change ?

vincentbernat:
Done. I have also appended a sysusers file to create the user. I didn't modify the Debian packaging because we don't have that yet. I am unsure if recent Fedora has switched to it yet and I didn't find the appropriate RPM macros for it. I suppose that some Redhat/Fedora user may come with the appropriate snippet when they hit the issue.

thomas-mangin:
You know I trust you to merge your work :p

vincentbernat:
OK ;-)

vincentbernat:
In other commits, I changed ETC to `/etc` instead of `/etc/exabgp`. However, I notice that the problem is also present in the SysV init script and that in the Travis script, there is also `ETC` set to `etc/exabgp`. So, I am a bit lost. In `application/bgp.py`, folder is set to `ETC`, `exabgp` and `envfile` to `folder`, `exabgp.env`, hence my thinking of setting `ETC` to /etc. However, I also see:

```python
	if not os.environ.get('ETC',''):
		os.environ['ETC'] = folder
```

So, now, when `ETC` is not set, it is set to `/etc/exabgp` but when it is set, it stays at `/etc`.

Maybe I should have opened an issue :)

vincentbernat:
Oh, in master, that's not the same thing!

vincentbernat:
So, the good option for both 3.4 and master would be to drop ETC and use `--folder`, right?

thomas-mangin:
@vincentbernat I changed this from env to --folder and can not fully recall why .. I believe I wanted to not use "global" env. variable as "ETC" is not really unique. As I do not really deal with installation ( I use ExaBGP from source and use absolute paths ) I am open to suggestion ( even radical as we are working towards version 4.0 )

thomas-mangin:
The initial reason exabgp configuration was in a folder called etc/exabgp were as ExaBGP is configured using a configuration file and the environment files, so it made sense to have a folder for these which could also be used for the helper scripts ( as it is still the case in the repo in etc/exabgp )

I did not write the initial debian package so I just accepted the work as-it. So it may be that it should be changed as my own script in sbin/exabgp does use etc/exabgp.

vincentbernat:
I am completely lost by your comment. :)

What is in 4.0 seems fine: not using `ETC`, preferring folder. So, it should be OK to use `--folder /etc/exabgp` everywhere (in packaging I mean).

Also, I don't have any problem with `/etc/exabgp`. It justs that in 3.4, if you set `ETC` to `/etc/exabgp`, it is like having `--folder /etc/exabgp/exabgp` and `--env /etc/exabgp/exabgp/exabgp.env`. However, at some other places, `ETC` is used as the relative base for processes, so if we set it to `/etc`, we have `--folder /etc/exabgp`, `--env /etc/exabgp/exabgp.env` (good) but processes are searched in `/etc` (bad).

Using `--folder` is OK because we get `--env /etc/exabgp/exabgp.env` and processes are searched in `/etc/exabgp` (because `ETC` is set to `folder`).

thomas-mangin:
Please have a look at the patch ( and feel free to change things if required )

vincentbernat:
I agree, this is mostly what I have done in #320, except I didn't touch the non-Debian bits.

vincentbernat:
I have pushed another commit on top of yours. I think this is OK. I'll close the other PR.

thomas-mangin:
thanks - with the latest patch I sent it should be completed.

PR #320, opened 2015-08-19
init: don't use ETC, use --folder instead

vincentbernat:
This should be compatible with both 3.4 and master.

It seems to be the best way to be consistent with 3.4 and master. See #319 for some explanations on the problems of ETC in 3.4.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/320)


vincentbernat:
Already handled in #319

PR #322, opened 2015-08-19
init: let ExaBGP drop privileges itself

vincentbernat:
We don't know exactly what kind of privileges ExaBGP may need at
start. It is likely to have to bind to port 179 and therefore requires
root privileges. systemd and SysV init script shouldn't drop
privileges. ExaBGP should do it after initialization.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/322)


vincentbernat:
Sorry for this back and forward. :)

thomas-mangin:
LOL - please never look at my commit history then :laughing:

Issue #328, opened 2015-09-02
Questions regarding Graceful Restart


charlescng:
Our organization has been using ExaBGP in our internal networking peering with our Arista equipment. Generally speaking, things have been working smoothly as expected but we do have issues with Graceful Restart. It’s not working as we are expecting and we like to get more information.

There have been cases where a virtual IP is blackholed or moved to a route with a lower local-preference for a few seconds during an ExaBGP restart. We have seen the following during a packet capture:

1. Graceful restart was successfully negotiated. No `NOTIFICATION` message was sent by ExaBGP before BGP connection termination.
2. The Arista switch peering with ExaBGP is correctly retaining the stale routes when ExaBGP terminated the BGP connection.
3. ExaBGP reestablishes the BGP connection. Since it has no RIB entries, it sends an `EOR` as required as per RFC 4724 (Section 4.1)
4. The switch peer drops all stale routes after the `EOR` was received as per RFC 4724 (Section 4.2)
5. The child process of ExaBGP sends a command to ExaBGP to announce routes
6. The original routes are re-established

Step 4 makes sense as ExaBGP does not persist its RIB to disk (https://github.com/Exa-Networks/exabgp/issues/45). There is no way for ExaBGP to resend the stale routes before the mandatory `EOR` message. At this point, the routes are lost and thus causes a momentary lost of routes to the host.

How does Graceful restart work in this context? Have we configured ExaBGP incorrectly?

Our ExaBGP config looks like this:
```
group neighbours {
  group-updates;

  process MyProcess {
    receive {
      neighbor-changes;
    }
    encoder json;
    run “/usr/local/bin/MyProcess";
  }

  neighbor 192.168.1.2 {
    router-id router1;
    local-address 192.168.1.4;
    local-as 65534;
    peer-as 65001;
    graceful-restart;
  }

  neighbor 192.168.1.3 {
    router-id router2;
    local-address 192.168.1.4;
    local-as 65534;
    peer-as 65002;
    graceful-restart;
  }
}
```

@thomas-mangin

thomas-mangin:
Hi Charles - thank you for this very good report. 

For the absence of the NOTIFICATION message. I suggest that we use a new issue (or private mail) to track this particular question as I will need to understand how the session was tear down to see if this behaviour was expected or not.

[...]

PR #333, opened 2015-09-21
Add the 'manual-eor' configuration to not send an initial EOR

charlescng:
Add a neighbour configuration `manual-eor` so that we can optionally not send an EOR in the initial `UPDATE` when establishing a BGP connection.

This follows a discussion in an issue:
```
It would simply add an option per peer to not send the EOR and then have the process always send it ( which make sense as in your scenario, it is unlikely that the configuration will contain any routes at all ).
The logic would then be moved into the helper process.
```
https://github.com/Exa-Networks/exabgp/issues/328

I tested this in our test environment and it resolved our route disruption issue when we restart ExaBGP. We're currently in production with ExaBGP 3.4.x so the changes are for the `3.4` branch. Another PR for the master branch will be created.

Thanks for the help @thomas-mangin to get this working. It's my first 'real' Python experience.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/333)


charlescng:
I'm not sure how to fix the coverage failure:
`coverage.misc.CoverageException: Unrecognized option '[report] exclude_line=' in config file .coveragerc`

PR #334, opened 2015-09-21
Add the 'manual-eor' configuration to not send an initial EOR

charlescng:
These are the same changes as https://github.com/Exa-Networks/exabgp/pull/333 for the `master` branch.

We don't currently use the 3.9 code base and our production configuration files are not compatible. The test case pass but I have not test this outside of the `3.4` branch.

@thomas-mangin


charlescng:
Just like the 3.4 branch PR, I'm not sure what caused the coverage test to be broken:
`coverage.misc.CoverageException: Unrecognized option '[report] exclude_line=' in config file .coveragerc`

PR #341, opened 2015-09-29
Corrected call to TTL()

borjam:
The TTL function was called from the established() method, which is called after the socket is connect()ed. 

At least on FreeBSD it doesn't work, would need testing on other platforms. Calling TTL() *before* the connect() is made, works.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/341)


thomas-mangin:
Hello again - same here why is this PR closed too ?

borjam:
Reopened. Sorry for the noise.

PR #344, opened 2015-10-01
cleanup NLRI code wrt bagpipe-specific things

tmmorin:
* revert previous change: have nexthop as a string for IPVPN.new

Bagpipe now subclasses IPVPN, so:
* remove eq  (not used by exabgp)
* remove \__hash\__ (which was inconsistent with \__eq\__/\__neq\__ anyways)

### DIFF
--- a/lib/exabgp/bgp/message/update/nlri/inet.py
+++ b/lib/exabgp/bgp/message/update/nlri/inet.py
@@ -48,13 +48,6 @@ class INET (NLRI):
 			self.path_info == other.path_info and \
 			self.nexthop == other.nexthop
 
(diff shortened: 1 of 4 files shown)


thomas-mangin:
Thank you - merged.

thomas-mangin:
It may have broken your test code tho - it is still using a NextHop object I  believe .. ?

tmmorin:
Nope. The test code is not initiating the nexthop (relying on the default value), since its not impacting pack/unpack anyways.

tmmorin:
Thanks for merging btw !

PR #352, opened 2015-11-21
Bug fix: ConditionPathExists is a systemd.unit directive

johanfleury:
Hi.

According to [systemd documentation](http://www.freedesktop.org/software/systemd/man/systemd.unit.html), `ConditionPathExists` must in the `[unit]` section, not in the `[service]` section.

This is the log systemd generate: `[/lib/systemd/system/exabgp.service:9] Unknown lvalue 'ConditionPathExists' in section 'Service'`.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/352)

### DIFF
--- a/etc/systemd/exabgp.service
+++ b/etc/systemd/exabgp.service
@@ -1,10 +1,10 @@
 [Unit]
 Description=ExaBGP
 After=network.target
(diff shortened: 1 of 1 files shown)


vincentbernat:
Thanks for the fix!

PR #363, opened 2016-01-04
a link to ExaZK

shtouff:
Hello,

Here at BlaBlaCar, we use ZooKeeper as a directory for our services. We also are used to run ExaBGP with healthcheck to announce our VIPs on our BGP infrastructure, on our load-balancers.

We created a script: exazk, to directly connect exabgp to ZK, and this way control what exabgp will announce, directly from ZK.

We were greatly inspired by healthcheck, off course, and I though it could be of some interest for the community of ExaBGP users.

Regards,

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/363)


thomas-mangin:
Thank you very much for sharing this work with the community. Well, blablacar is all about sharing after all :stuck_out_tongue_winking_eye:

ton31337:
I'm using this too already few months. http://github.com/ton31337/exazk

PR #364, opened 2016-01-06
Add ExaZK as use case

ton31337:
Add ExaZK as use case.
/cc @thomas-mangin

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/364)

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ Please let us know if you use it.
  * [ExaBGPmon](https://github.com/thepacketgeek/ExaBGPmon) web interface to ExaBGP
  * [ERCO](https://erco.xyz/) web interface to ExaBGP
  * [ExaZK](https://github.com/shtouff/exazk) a plugin to interface ExaBGP & ZooKeeper
+ * [exazk](https://github.com/ton31337/exazk) a ruby solution to interface ExaBGP & ZooKeeper to achieve service HA
 
 ## Self Promotion
 


thomas-mangin:
Hi Donatas, happy to add your solution to the list ( i believe it is already on the wiki )
As both yours an Blablacar have both the same name, it would be nice to be able to present the difference between both in the summary text.

ton31337:
@thomas-mangin is it better? Actually this is in development stage, but we are using it already, so many updates will be pushed soon in 1-2 months.

thomas-mangin:
I was more wondering how a reader could decided what was best for them between the two solution ? Are they working the same way ?

ton31337:
@thomas-mangin as I understand the principle is the same as with previous one, except that this is written in ruby language and more lightweight (it's just one file solution).

thomas-mangin:
what about changing the text like this ?
  * [ExaZK](https://github.com/shtouff/exazk) a python solution to interface ExaBGP & ZooKeeper to achieve service HA
 + * [exazk](https://github.com/ton31337/exazk) a ruby solution to interface ExaBGP & ZooKeeper to achieve service HA

ton31337:
@thomas-mangin :+1:

thomas-mangin:
Great . Do you want to update your commit request ? Or I can do it - as you with.

thomas-mangin:
Sorry - done.

PR #368, opened 2016-01-21
RTRecord support

tmmorin:
These two commits add support for RTRecord (as specified in draft-fm-bess-service-chaining).

The first commit simply make RouteTarget easily extensible (does not change any behavior in ExaBGP).

The second commit adds support for RTRecord based on the first commit.
If preferred, because the spec still is an individual draft and the subtype not IANA-registered, the second commit can be left apart (we can add/register it from bagpipe-bgp).

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/368)

### DIFF
--- a/lib/exabgp/bgp/message/update/attribute/community/extended/rt_record.py
+++ b/lib/exabgp/bgp/message/update/attribute/community/extended/rt_record.py
@@ -1,0 +1,33 @@
+from exabgp.bgp.message.update.attribute.community.extended import ExtendedCommunity
+from exabgp.bgp.message.update.attribute.community.extended import rt
+
+import logging
+log = logging.getLogger(__name__)
+
+# draft-fm-bess-service-chaining
+
+
+class RTRecord(rt.RouteTarget):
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Very good patches.

lib/exabgp/bgp/message/update/attribute/community/extended/rt_record.py
```diff
@@ -0,0 +1,33 @@
+from exabgp.bgp.message.update.attribute.community.extended import ExtendedCommunity
+from exabgp.bgp.message.update.attribute.community.extended import rt
+
+import logging
```
thomas-mangin: Just noticed this - should I just remove these two lines ?

tmmorin: Ah yes, correct, the "import logging" and the line above are not needed.

Thanks!

thomas-mangin: Ok, done with some other cleanup.

PR #369, opened 2016-01-30
fix neighbor address in show_routes

wtucker:
Output erroneously uses local_address instead of peer_address.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/369)


thomas-mangin:
Thank you.

PR #371, opened 2016-02-05
fix RTC wildcard route encoding/decoding

tmmorin:
[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/371)


thomas-mangin:
Thank you for this work.

PR #374, opened 2016-02-08
Update rpm spec file and add patch to build healthcheck

dzredding:
* The specfile attempted to package healthcheck which was not built or installed. I've included a patch for the specfile to process which builds healthcheck.
* The path that the specfile attempted to copy to /etc/ was nonexistent. The /etc/exabgp as provided contains content questionable for /etc, but I've fixed the specfile to copy what's provided. This should probably contain a dummy, commented config or be empty.
* The specfile failed to package /usr/share/exabgp.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/374)

### DIFF
--- a/redhat/healthcheck.patch
+++ b/redhat/healthcheck.patch
@@ -1,0 +1,12 @@
+diff --git a/setup.py b/setup.py
+index dc82683..d758aec 100644
+--- a/setup.py
++++ b/setup.py
+@@ -346,6 +346,7 @@ setup(
+        entry_points={
+                'console_scripts': [
+                        'exabgp = exabgp.application:run_exabgp',
++                       'healthcheck = exabgp.application:run_healthcheck',
+                ],
+        },
+ )
--- a/redhat/python-exabgp.spec
+++ b/redhat/python-exabgp.spec
@@ -3,13 +3,14 @@
 
 Name:           python-exabgp
 Version:        3.4.11
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        The BGP swiss army knife of networking (Library)
 
 Group:          Development/Libraries
 License:        BSD
 URL:            http://pypi.python.org/pypi/exabgp/
 Source0:        https://github.com/Exa-Networks/exabgp/archive/%{version}/exabgp-%{version}.tar.gz
+Patch0:		healthcheck.patch
 BuildArch:      noarch
 Provides:       exabgp-libs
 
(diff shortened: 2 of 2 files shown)


thomas-mangin:
The change could be incorporated upstream - will look into it.

vincentbernat:
We discussed the change in the past, but we did not do it because "healthcheck" was quite generic. We said that a user could just use `python -m exabgp healthcheck` instead.

thomas-mangin:
I remember the discussion now - thank you @vincentbernat - yes this could cause issues.

We could consider having "exabgp --run healthcheck" as an alternative.. and put some wording at that effect in the --help ? 

@dzredding what do you think ?

dzr0001:
@thomas-mangin I think that would work. I could resubmit a PR that removes all of the bits for packaging healthcheck rather than actually building it like I've done here.

dzredding:
@thomas-mangin That would probably work. I could submit a new PR to remove the healthcheck packaging rather than build healthcheck like I've done here.

thomas-mangin:
Hi, I am confused about your last post ...

dzredding:
@thomas-mangin So the changes I made in this PR addressed 2 problems.

1. The spec file expected the exabgp-healthcheck pieces to be in place. setup.py didn't build this, so I created a patch that the specfile uses to have setup.py build and install exabgp-healthcheck, with the intention that the patch would only be used for RPM installations.

2. The specfile didn't properly handle /usr/share* items at all. I added support for this.

Regarding item 1, I may have misinterpreted previous comments in a way that I thought it was undesirable to build exabgp-healthcheck. As such, I offered to make changes to remove the support for exabgp-healthcheck from the specfile rather than patch in support to build it as I've done in this PR.

thomas-mangin:
Make sense now :-) I will gladly take the /usr/share fix.

dzredding:
@thomas-mangin Ok, should I close this and just open new PR that addresses the /usr/share packaging only? 

Also, I failed to mention that I also addressed the lack of sysconfdir in the buildroot. The etc/exabgp files duplicate a lot of what's in /usr/share/exabgp, and I think having an uncommented default config might be dangerous. Do you have any thoughts on what you'd prefer to see there? I think either an empty directory or a config file with comments pointing to /usr/share/exabgp examples would be sufficient.

thomas-mangin:
Please leave it open so I do not forget to add the --run option.

thomas-mangin:
A new PR for the other feature would be good.

thomas-mangin:
--run implemented.

PR #375, opened 2016-02-15
fix wildcard RTC route len encoding + unit tests

tmmorin:
- len should be zero and not one for wildcard RTC routes
- add unit tests for wildcard RTC routes

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/375)


thomas-mangin:
Thank you Thomas

PR #386, opened 2016-03-10
Healthcheck fixes

benagricola:
@thomas-mangin as discussed in gitter, this fixes the issue with ip addr show parsing in the healthcheck script expecting whitespace at the end of the address line. This does not always necessarily exist, and is unimportant for matching on anyway because the only needed information is prior to this expected whitespace.

This patch also adds a `remove_ips()` method which is triggered on SIGTERM to remove any IP addresses that were added by `setup_ips()`. Since the healthcheck script must be running as root to remove the IP addresses on termination, the method captures `ip addr del` command failure to warn the user that it was unable to remove the IP address.

It may be better to check whether we're running as a privileged user and simply not run the ip addr del command, but I couldn't find a simple way to verify if the script is running privileged or not.

Please let me know if you'd like these commits squashed since this currently contains a reverted commit, and then a revert of that reverted commit to make this branch suitable for merging into master :)

---
This change is [Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/386)

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -265,7 +265,7 @@ def loopback_ips(label):
         # Use "ip" (ifconfig is not able to see all addresses)
         ipre = re.compile(r"^(?P<index>\d+):\s+(?P<name>\S+)\s+inet6?\s+"
                           r"(?P<ip>[\da-f.:]+)/(?P<netmask>\d+)\s+.*")
-        labelre = re.compile(r".*\s+lo:(?P<label>\S+)\s+.*")
+        labelre = re.compile(r".*\s+lo:(?P<label>\S+).*")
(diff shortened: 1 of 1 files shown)


thomas-mangin:
@vincentbernat LGTM - do you want to check and if happy merge ?

vincentbernat:
`os.getuid() == 0` will tell you if you are running as root. I don't think this should be checked but I don't have a definitive opinion on this. Otherwise, LGTM.

benagricola:
My personal opinion is that removing the IP is less important than failing to add the IP and exiting with an error :smiley: - I think as long as we log the fact that the IP couldn't be removed due to privileges and that the healthcheck needs to run as root for that to be effective then we're all good.

vincentbernat:
Could you squash everything?

Issue #387, opened 2016-03-14
ttl_security should allow 255


causey:
Hello,

We use a ttl_security value of of 255 in our configuration, which is incompatible with this validation:

https://github.com/Exa-Networks/exabgp/blob/master/lib/exabgp/configuration/neighbor/parser.py#L85

Can I just update this value to be 256 and submit a PR?
Thanks!

Matt

causey:
Here's the related PR:

https://github.com/Exa-Networks/exabgp/pull/388

thomas-mangin:
Thank you - fixed with another patch - help much appreciated.

PR #388, opened 2016-03-15
Allowing a ttl-security of 255

causey:
From related issue: https://github.com/Exa-Networks/exabgp/issues/387

We use a ttl_security value of of 255 in our configuration, which is incompatible with this validation.  This change just allows a value of 255.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/388)


thomas-mangin:
Thank you. I have changed the test to be > and not >= ... fixed

causey:
Thanks!  Do you know when's the next release?  I'm curious when this'll make it to pypi.

thomas-mangin:
The master branch is not going to see a release before a few months at best - the 3.4 release sees upgrade every few months when bug get fixed or backward compatible features are added.

PR #402, opened 2016-04-06
update so umask is configurable rather than hard-set in daemon.py

blegear:
alternative would be to omit umask altogether and have calling process filter that through
i've done it this way for minimal change in current behaviour

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/402)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -120,6 +120,8 @@ Version 4.0.0
    requested by: Borja Marcos
  * Feature: md5-ip allows to override local-address for the MD5 calculations
    requested by: Bryan Benson
+ * Feature: allow setting process umask via exabgp.daemon.umask
+   patch by: Bryan Le Gear
 
 Version 3.4.10
  * Fix: Fix parsing attributes with PARTIAL flag set
--- a/debian/exabgp.upstart
+++ b/debian/exabgp.upstart
@@ -5,7 +5,6 @@ stop on runlevel [!2345]
 
 console none
 respawn
-umask 022
 reload signal USR1
 
 env PATH=/sbin:/usr/sbin:/bin:/usr/bin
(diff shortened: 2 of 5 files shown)


thomas-mangin:
Hello @blegear,

This patch seems fine. As I have not tested can you confirm that you are able to generate the ini file with "exabgp --fi" and that if saved it loads fine.

I will just ask @vincentbernat if the umask change in the debian file for upstart is fine with him (as he maintains this part) - if so I have no objection with the feature and will merge.

You may want to add an entry in the CHANGELOG announcing the feature too please.

blegear:
Have updated the PR. hadn't appreciated the env writer aspect when i first dived in but that now works using exabgp --fi.. I've put this through a python setup.py bdist-rpm / fresh deploy etc and this works.

WRT @vincentbernat / debian changes you mention, as far as i can tell setting umask in upstart / systemd would have had no appreciable affect as daemon.py sets this statically as soon as it runs (so my change should not alter behaviour vs previously) .

thomas-mangin:
Ok - thank you. Then why set it up as it is going to be overwritten anyway ?

thomas-mangin:
I will merge @vincentbernat can revert if required :-)

blegear:
thanks thomas! one more diff on this I'm afraid, editor expanded tabs -> spaces which i only caught post-commit.
Will raise one final PR for you to merge. :(

thomas-mangin:
Thanks - this is indeed an issue.

vincentbernat:
It was part of the contribution of the upstart job from @pyke369. It's not common to have umask set in the upstart job, maybe there was a reason for that or maybe it was just a left-over of a copy/paste from another project.

PR #404, opened 2016-04-06
IPSelf has no .index()

blegear:
eg:if old_change.attributes.index() == change.attributes.index() and old_change.nlri.nexthop.index() == change.nlri.nexthop.index():
AttributeError: 'IPSelf' object has no attribute 'index'
not certain if this is the right thing to do although it does seem to work in my environment. this diff is more to point out the issue i've seen and to suggest a possible fix

Heres a traceback
```
File "/usr/bin/exabgp", line 9, in <module>
load_entry_point('exabgp==4.0.0', 'console_scripts', 'exabgp')()
File "/usr/lib/python2.7/site-packages/exabgp/application/__init__.py", line 12, in run_exabgp
main()
File "/usr/lib/python2.7/site-packages/exabgp/application/bgp.py", line 246, in main
run(env,comment,configurations)
File "/usr/lib/python2.7/site-packages/exabgp/application/bgp.py", line 280, in run
ok = Reactor(configurations).run()
File "/usr/lib/python2.7/site-packages/exabgp/reactor/loop.py", line 207, in run
action = peer.run()
File "/usr/lib/python2.7/site-packages/exabgp/reactor/peer.py", line 700, in run
r = direction.generator.next()
File "/usr/lib/python2.7/site-packages/exabgp/reactor/peer.py", line 683, in _run
self._reset(direction)
File "/usr/lib/python2.7/site-packages/exabgp/reactor/peer.py", line 220, in _reset
self.neighbor.rib.reset()
File "/usr/lib/python2.7/site-packages/exabgp/rib/__init__.py", line 38, in reset
self.outgoing.reset()
File "/usr/lib/python2.7/site-packages/exabgp/rib/store.py", line 43, in reset
for update in self.updates(True):
File "/usr/lib/python2.7/site-packages/exabgp/rib/store.py", line 225, in updates

and the config im running.
process dynamic-route-updater {
  encoder text;
  run /usr/local/bin/exabgp-myroute --config /etc/exabgp/exabgp-myroute.conf --logfile /var/log/exabgp/exabgp-myroute.log;
}

neighbor 10.220.130.1 {
  router-id 10.220.130.37;
  local-address 10.220.130.37;
  local-as 65022;
  peer-as 65505;

  capability {
    graceful-restart;
  }
  api {
    processes [ dynamic-route-updater ];
  }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/404)

### DIFF
--- a/lib/exabgp/rib/store.py
+++ b/lib/exabgp/rib/store.py
@@ -222,7 +222,7 @@ class Store (object):
 						if nlri_index in announced:
 							old_change = announced[nlri_index]
 							# it is a duplicate route
-							if old_change.attributes.index() == change.attributes.index() and old_change.nlri.nexthop.index() == change.nlri.nexthop.index():
+							if old_change.attributes.index() == change.attributes.index() and old_change.nlri.nexthop == change.nlri.nexthop:
 								continue
 					elif change.nlri.action == OUT.WITHDRAW:
 						if nlri_index not in announced:


thomas-mangin:
Hello @blegear,

Thank you for this patch, but this is not the right solution. change.attributes should not be an `IPSelf` but an `Attributes` class.

If you run exabgp with the option "-d -p", it will drop you into the debugger when the issue occurs (in case you ever want to debug things yourself). I will need to look at how the API can generate a route I believe is not right.

Could you please confirm the format of the route you are using through the API ?

Thomas

blegear:
Thanks for quick response on this.. To add a bit more info.. In the simplest case, exabgp-myroute process is announcing only 1 route and it is of exactly this form, no updates are consumed. This is enough to reproduce.
"announce route 10.220.254.22/32 next-hop self med 100"

This was working in 3.4.11 , after upgrade to latest master branch (+ translate config to newer format) it breaks, if this is just a matter of changing the format of my advertisement i'm more than happy to do so, although as you say it'd be worth fixing as shouldn't have been of type IPSelf in the first place.

Behaviour is such that the first route update is correctly processed by exabgp and is communicated to next-hop router which sees prefix in its own routes received, but the process exits with the stack trace i provided shortly (maybe a second or two) afterwards, then peering is then dropped.

if you need any more info just let me know.

thomas-mangin:
Ok - looking at a fix now..

thomas-mangin:
Very surprised as this part of the code is tested (using the latest master)

TERMINAL 1 > SERVER=G ./qa/bin/conversation
TERMINAL 2 > CLIENT=G ./qa/bin/conversation

```
process announce-routes {
	run ./api-nexthop-self.run;
	encoder json;
}

neighbor 127.0.0.1 {
	router-id 1.2.3.4;
	local-address 127.0.0.1;
	local-as 1;
	peer-as 1;
	group-updates false;

	family {
		ipv4 unicast;
	}

	api {
		processes [ annouce-routes ];
	}
	static {
		route 10.0.0.1 next-hop self;
		route 10.0.0.2 {
			next-hop self;
		}
	}
}
```

```
#!/usr/bin/env python

import sys
import time

messages = [
'neighbor 127.0.0.1 announce route 1.1.0.0/16 next-hop self',
'announce route 2.2.0.0/16 next-hop self',
]

while messages:
	message = messages.pop(0)
	sys.stdout.write( message + '\n')
	sys.stdout.flush()
	time.sleep(0.1)

while True:
	time.sleep(1)
```

thomas-mangin:
Ok - the test fails but it is not reported as failing .. sigh !

benagricola:
Running into this issue too.

This happens when a duplicate (already-announced) route is re-announced. the attributes checks work fine, but the next-hop check causes an unhandled exception due to the aforementioned lack of an `index()` method of `IPSelf`.

I think the fix in the PR is incorrect, but not because of the reason @thomas-mangin stated above:

> Thank you for this patch, but this is not the right solution. change.attributes should not be an IPSelf but an Attributes class.

The reason this isn't correct is that using `.nexthop` directly will cause python to compare the `__repr__()` of the `IPSelf` object, which is always the string `self`. This means a next-hop-self AFI change would not be identified as a route change.

I reckon this should probably be handled as a third case (if `IPSelf`, compare AFI values).

If this sounds good, I'll sort out a PR.

thomas-mangin:
I looked into it quite some time ago, but the class issue is real (but it may not be  the only way to fix the issue). Happy to consider another PR.

thomas-mangin:
@benagricola - sorry for the delay in answering - I have been busy.

I support the approach in your tree : https://github.com/benagricola/exabgp/commits/fix-duplicate-ipself-announce

However:
```
(Pdb++) AFI.names[self.afi]
'ipv4'
(Pdb++) '.'.join([str(ord(_)) for _ in AFI.names[self.afi]])
'105.112.118.52'
```
So we need the index function to return something like:
`return 'fake-' + AFI.names[self.afi]`

If you update the patch that way, I will merge it.

thomas-mangin:
Implemented the patch as proposed - closing.

Issue #426, opened 2016-06-08
IPv6


azlux:
Is the IPv6 working with ExaBGP ? I have this message (it's a clone of the master branch)
````
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | processes     | Command from process service-dynamic : announce flow route {\n match {\n source 2a01:728:101:11:594b:bf07:136c:79c9/64;\n  protocol 17;\n }\n then {\n discard;\n }\n }\n
Wed, 08 Jun 2016 15:01:58 | WARNING  | 24093  | reactor       | callback | handling 'announce flow route {\n match {\n source 2a01:728:101:11:594b:bf07:136c:79c9/64;\n protocol 17;\n }\n then {\n discard;\n }\n }\n' with announce_flow
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | reactor       | callback | installing announce_flow
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | reactor       | callback | running
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | configuration | > route            |
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | configuration | > match            |
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | configuration | . source           | '2a01:728:101:11:594b:bf07:136c:79c9/64'
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | configuration |
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | configuration | syntax error in api command route/match
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | configuration | line 3: source 2a01:728:101:11:594b:bf07:136c:79c9/64 ;
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | configuration |
Wed, 08 Jun 2016 15:01:58 | INFO     | 24093  | configuration | invalid literal for int() with base 10: '2a01:728:101:11:594b:bf07:136c:79c9'
Wed, 08 Jun 2016 15:01:58 | WARNING  | 24093  | reactor       | command could not parse flow in : announce flow route {\n match {\n source 2a01:728:101:11:594b:bf07:136c:79c9/64;\n protocol 17;\n }\n then {\n discard;\n }\n }\n
[...]

PR #428, opened 2016-06-09
ipv6 parsing

azlux:
try to solve #426 issue.
I hope the default offset is 0

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/428)

### DIFF
--- a/lib/exabgp/configuration/flow/parser.py
+++ b/lib/exabgp/configuration/flow/parser.py
@@ -50,10 +50,14 @@ def flow (tokeniser):
 
 def source (tokeniser):
 	data = tokeniser()
-	if data.count('/') == 1:
+	if data.count('.') == 3 and data.count(':') == 0:
 		ip,netmask = data.split('/')
 		raw = ''.join(chr(int(_)) for _ in ip.split('.'))
 		yield Flow4Source(raw,int(netmask))
+	elif data.count('/') == 1:
+		ip,netmask = data.split('/')
+		offset = 0
+		yield Flow6Source(IP.pton(ip),int(netmask),int(offset))
 	else:
 		ip,netmask,offset = data.split('/')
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Thank you very much for the patch, much appreciated. I can however not merge it as you have have some IPv6 in the form ( like :ffff:192.0.2.128 ), so the first test need to test for the presence of : and not .

azlux:
I can do it for the specific format. But 0000:0000:0000:0000:0000:0000:127.127.127.127 is a [depreciate format](https://tools.ietf.org/html/rfc4291#section-2.5.5.1) now.
Do you really want to use it ?

lib/exabgp/configuration/flow/parser.py
```diff
@@ -50,22 +50,30 @@ def flow (tokeniser):
...
 
 
 def destination (tokeniser):
 	data = tokeniser()
-	if data.count('/') == 1:
+	if data.count('.') == 3 and data.count(':') == 0 :
 		ip,netmask = data.split('/')
 		raw = ''.join(chr(int(_)) for _ in ip.split('.'))
 		yield Flow4Destination(raw,int(netmask))
 		return
```
thomas-mangin: If you can please remove that return - I will merge :-)

thomas-mangin:
Also .... sorry ... once the return is gone can you please merge all the commit in one. It can be done with git rebase -i HEAD~5  using "s" for squashing the commit together (keep the first one and use s for the following).

azlux:
first time I use the remake function.Hope it's work.
What do you mean "the return" into your previous comment ? (I've suppose you was talking about the additional space)

thomas-mangin:
rebase - just gave you the syntax to run after another commit
return - there is/was a left over useless line with only a "return" in the code

thomas-mangin:
Thank you very much for the patch.

PR #432, opened 2016-06-21
FSM state fix

tmmorin:
Transitioning to ACTIVE should only be done if the connection attempt fails.
During the TCP connection attempt, the correct state is CONNECT.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/432)


thomas-mangin:
Thomas, I am merging but I need to rewrite the whole FSM - it is currently wrong.

PR #439, opened 2016-06-28
Fix AS4_AGGREGATOR attribute handling

s2ugimot:
Fix to set proper AGGREGATOR and AS4_AGGREGATOR attributes as written in RFC 4893

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/439)

### DIFF
--- a/lib/exabgp/bgp/message/update/attribute/aggregator.py
+++ b/lib/exabgp/bgp/message/update/attribute/aggregator.py
@@ -42,7 +42,10 @@ class Aggregator (Attribute):
 		if negotiated.asn4:
 			return self._attribute(self.asn.pack(True)+self.speaker.pack())
 		else:
-			return self._attribute(self.asn.trans()+self.speaker.pack()) + self._attribute(self.asn.pack(True)+self.speaker.pack())
+			if self.asn.asn4():
+				return self._attribute(self.asn.trans()+self.speaker.pack()) + Aggregator4(self.asn,self.speaker).pack(negotiated)
+			else:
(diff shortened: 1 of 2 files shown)


a16:
+1
I've tested exabgp with this changes. I believe it works fine in that generation of AGGREGATOR/AS4_AGGREGATOR.

What do you think Thomas?

thomas-mangin:
Konbanwa @a16 @s2ugimot 

As I have no router with ASN4 enabled, I can not test the code. 
I would have expected 
`return self._attribute(self.asn.trans()+self.speaker.pack()) + Aggregator4(self.asn,self.speaker).pack(negotiated)`
to be something like 
`return self._attribute(self.asn.trans()+self.speaker.pack()) + self._attribute(Aggregator4(self.asn,self.speaker).pack(negotiated))`
what do you think ?

thomas-mangin:
if you tested both the case of ASN4 session and non-ASN4 session - happy to keep the patch in.

s2ugimot:
@thomas-mangin Hello, and ohayo gozaimasu :)
Thank you for merging. As @a16 mentioned already, we've tested all possible patterns with IOS-XR/JUNOS/GoBGP and all seems fine.

thomas-mangin:
I have tweaked your patch https://github.com/Exa-Networks/exabgp/commit/8f5c6d11208aa366282ee4d1b299409c2586262d and backported it to 3.4 https://github.com/Exa-Networks/exabgp/commit/caf423424c87234403de3c277dd97099472b23ef

Once again Arigatou gozaimashita - your help is much appreciated.

PR #450, opened 2016-07-19
Extract method to return ordered rules string for external use

tmmorin:
This does not fix or improve exabgp, but simply makes code reuse easier in bagpipe-bgp.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/450)


thomas-mangin:
can you please rename this method to _flow_rules. While this is fine for bagpipe to have such feature, it should be clear to the reader that this is not really a supported API. Also ExaBGP does not have a convention of using str_ for functions.

Perhaps the code should break down this even more - I may create a sub-helper function for this as ultimately I do not really like the way the string are made. 

I should really create a list of tuple (name, value) and have the code check if the value is a list and render it accordingly. Currently it is messy (but works).

thomas-mangin:
finally renamed to _rules

PR #454, opened 2016-08-01
stop exiting when started without neighbors configured

### LINKED ISSUE #372: allow exabgp to start without any neighbours
It would simplify our orchestration if we could start a fixed amount of instances on a host without any neighbours configured.
Over time neighbours would be added and exabgp instructed to reload its configuration.
Currently exabgp just exits.
Thanks,
Florian

jordangedney:
Fixes #372 

The shutdown, reload, restart, and process scheduler were unnecessarily
tied into a self.peers check in the main run-loop- this change pulls them out
of this check. (And removes seemingly dead code at 227.)

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/454)

### DIFF
--- a/lib/exabgp/reactor/loop.py
+++ b/lib/exabgp/reactor/loop.py
@@ -179,26 +179,30 @@ class Reactor (object):
 
 		while True:
 			try:
-				while self.peers:
-					start = time.time()
-					end = start+self.max_loop_time
+				start = time.time()
+				end = start+self.max_loop_time
 
-					if self._shutdown:
-						self._shutdown = False
-						self.shutdown()
-					elif self._reload and reload_completed:
-						self._reload = False
-						self.load()
-						self.processes.start(self._reload_processes)
-						self._reload_processes = False
-					elif self._restart:
-						self._restart = False
-						self.restart()
-					elif self.route_update:
+				if self._shutdown:
+					self._shutdown = False
+					self.shutdown()
+					break
+				elif self._reload and reload_completed:
+					self._reload = False
+					self.load()
+					self.processes.start(self._reload_processes)
+					self._reload_processes = False
+				elif self._restart:
+					self._restart = False
+					self.restart()
+
+				ios = {}
+				keys = set()
+
+				if self.peers:
+					if self.route_update:
 						self.route_update = False
 						self.route_send()
 
-					ios = {}
 					keys = set(self.peers.keys())
 
 					while start < time.time() < end:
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Thank you very much for this patch. Much appreaciated. I was going to work on it today :-)

`if self.listener:` is currently under `if self.peers` which is wrong, we should handle incoming connection even if no peers are setup, only if to clear the sockets.

So if you un-indent that block, I will accept the patch 'as it'

thomas-mangin:
Could you please also forward port this patch to master too ? If not practical I will do it when I merge tho.

thomas-mangin:
(sorry, another issue) also, can you please move `start = time.time()` and `end =` at the start of the `try:` block - we want to upper bound the time it takes to process API command and message processing time. Checking for incoming connection should be quite fast, and is therefore ignored (should not but works and is simpler).

Also the self.ready block should always be run, so be moved down one level too.

thomas-mangin:
and finally (hopefully :p) ... could you take all the patches and just make one ( git rebase -i HEAD~<number of patch> may help, use 's' for squash for the second and following patches you want to merge into the first). Never done it for a feature request on github.

thomas-mangin:
I have not told you about the fact that you will need to write `keys = set()` if there is no peers to not have undefined variable and that in that case self.schedule() can move where it used to be ..

jordangedney:
I'm not sure I understand your last suggestion, probably based on a misunderstanding of "Also the self.ready block should always be run, so be moved down one level too."... 

I've pulled self.ready out of the while start < time.time() so that it will always run given peers. (It doesn't make sense for it to run without peers, as IOS seems to be set based on the peers... and without IOS self.ready doesn't do anything; sleep, but no change)

Like I said, I'm probably just misunderstanding; here is a haste to give you an idea of my current understanding: http://hastebin.com/iwotizusip.py

thomas-mangin:
in the `if not keys:` block, which was `if not self.schedule() and not keys:` your changed the ready line to `ready = self.ready(ios.keys(),end-time.time())` from `ready = self.ready(ios.keys() + self.processes.fds(),end-time.time())`

You should undo this change and have the `self.schedule()` where it was and keep the `self.processes.fds()`

It allows the reactor to sleep UP to reactor.sleep time (1 second by default) UNTIL we have network or forked process activity ...

This block should be un-indented and always run (same level as  `if self.peers:`) so in that case the variable `keys` may not be defined if we do not have peers, so making sure the variable is initialised to `set()` is important.

jordangedney:
Let me know if these changes are more what you had in mind- I'll pull against master if they are.

thomas-mangin:
Great, it all looks correct but will need to double check as I could have missing something.

PR #459, opened 2016-08-09
Fix JSON encoding of updates without NLRIs

dhammika:

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/459)

### DIFF
--- a/lib/exabgp/reactor/api/encoding.py
+++ b/lib/exabgp/reactor/api/encoding.py
@@ -319,10 +319,10 @@ class JSON (object):
 			remove.append(s)
 
 		nlri = ''
-		if not add and not remove:  # an EOR
-			if not update.nlris:
-				raise RuntimeError('no update.nlri: %s %s' % (type(update),dir(update)))
-			return update.nlris[0].json()
+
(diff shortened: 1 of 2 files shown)


lib/exabgp/reactor/api/encoding.py
```diff
@@ -319,10 +319,10 @@ def _update (self, update):
...
 
 		nlri = ''
-		if not add and not remove:  # an EOR
-			if not update.nlris:
-				raise RuntimeError('no update.nlri: %s %s' % (type(update),dir(update)))
-			return update.nlris[0].json()
+
+		if not add and not remove:
+			if update.nlris and update.nlris[0] == nlri.EOR:  # an EOR
+				return update.nlris[0].json()
```
thomas-mangin: would we not still need to raise if there is no update.nlris ?

dhammika: No, this is a case where we just get attributes without nlris.

thomas-mangin:
Merging, I notice that nlri.EOR is going to be an issue as nlri is string .. looking to fix.

PR #464, opened 2016-08-16
Fix api race conditions

brijohn:
There are several places where its possible for the neighbors to be modified by a configuration reload while in the middle of iterating over them. This should fix the ones I spotted. Some these appear to be already fixed in master, but not all.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/464)


thomas-mangin:
Thank you before I merge this one, can you update it and add an entry in the CHANGELOG please.

thomas-mangin:
I already merge the master one (as all bugs fixed in 3.4 are forward ported).

thomas-mangin:
Thank you.

PR #471, opened 2016-08-25
Update neighbor.py

doddt:
Modify output of 'show neighbors' so that it will be accepted by Configuration.reload(). This consists of removing colons after the keywords group-updates, auto-flush, adj-rib-out, and ttl-security.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/471)


thomas-mangin:
Thank you very much for that .. I am currently work (yes, Friday 23:00 still at work and it is likely to be this way for at least a few weeks). I will look and comment when I can.

thomas-mangin:
Thank you.

PR #472, opened 2016-08-25
Update neighbor.py

doddt:
Modify 'show neighbors' output so that it may be loaded in Configuration.reload(). This consists of removing the colons following the keywords group-updates, auto-flush, adj-rib-out, outgoing-ttl, and incoming-ttl.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/472)


thomas-mangin:
Not ignoring your work - but I am under some deadlines are work - hopefully I will be able to go back to a more normal live next week and merge this.

thomas-mangin:
Thank you make sense.

PR #479, opened 2016-09-01
Update community.py

doddt:
Add missing case for 'nopeer' community string when serializing communities.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/479)


thomas-mangin:
This branch should be used.

PR #481, opened 2016-09-03
Fix the broken flush route api

brijohn:
This fixes the flush route api so it will actually flush any unsent routes when auto-flush is disabled. 

Thomas, let me know if you would like me to  prepare a PR request for this against master as well. Same for the previous show neighbor status api PR.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/481)


thomas-mangin:
Hello Brian, Thank for very muck for this fix. I would gladly accept a patch against master with the same fix.

PR #492, opened 2016-09-09
An asn4 capable peer should be able to speak with an asn2 peer.

brijohn:
This section of code is breaking the ability of an asn4 capable peer talking with an asn2 only peer.  When removed things seem to work as expected per the rfcs, the OPEN message sets AS_TRANS as its two byte asn and UPDATE messages contain the AS4_PATH attribute containing the asn4 value.

Is there a reason this code is here that I am overlooking?

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/492)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -37,6 +37,8 @@ Version 3.4.17
     patch by: doddt
  * Change: syslog format changed to be in line with other application
     patch by: Brian Johnson
+ * Fix: Allow asn4 peer to speak with asn2 only peer
+    patch by: Brian Johnson
 
 Version 3.4.16
  * Feature: allow users to decide if processes must be run before or after we drop privileges
--- a/lib/exabgp/bgp/message/open/capability/negotiated.py
+++ b/lib/exabgp/bgp/message/open/capability/negotiated.py
@@ -109,14 +109,6 @@ class Negotiated (object):
 		# 		self.received_open_size = self.peer.bgp.received_open_size - 19
 
 	def validate (self, neighbor):
-		if not self.asn4:
-			if neighbor.local_as.asn4():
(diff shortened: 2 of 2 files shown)


thomas-mangin:
AFAICS the code says .. if we have an ASN4 and our peer does not speak ASN4 - do not establish a session.

You are right, that it is narrow minded :p 
It should be replaced by some code to fake our ASN to AS_TRANS .. removing the code is not the right patch tho, that can not be right or does it work ?

brijohn:
> AFAICS the code says .. if we have an ASN4 and our peer does not speak ASN4 - do not establish a session.

Yes that is my take on the code as well and matches what we are seeing in our test cases.

> It should be replaced by some code to fake our ASN to AS_TRANS .. removing the code is not the right patch tho, that can not be right or does it work ?

It does in fact look like it current does the correct thing if this validation check is simply removed.  Line 62 in lib/exabgp/bgp/message/open/__init__.p appears to be setting AS_TRANS in the open message for your 2 byte asn if your actual asn is greater then 2 bytes. Similarly the pack method in the ASPath class seems to handle adding the AS4Path attribute when necessary as well.  I am wondering if this is old code that was left in there and is no longer needed?

For reference I've attached a pcap file that was taken with the validation removed. exabgp is 10.8.22.75 and the remote peer is 10.8.9.26.
[2byteinterop_afterfix.pcap.gz](https://github.com/Exa-Networks/exabgp/files/466115/2byteinterop_afterfix.pcap.gz)

thomas-mangin:
Yes, I must have written the right logic and forgot to remove that.
Can you please add a CHANGELOG entry and I will merge the patch.

Could you also forward port the patch to master please ?

PR #495, opened 2016-09-15
Update check.py

doddt:
Since exabgp can now run without peers, we should also be able to parse configurations that do not specify peers.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/495)


thomas-mangin:
Can you please create a new 3.4.18 section and add a feature section in it for this patch :-)

doddt:
CHANGELOG updated. I hope I did it right. :-)

PR #504, opened 2016-09-29
config: allow 'connect' option also in 'neighbor' scope in config

solidgoldbomb:
The 'connect' config option is currently only allowed in the 'group' scope even though it is already inherited to be a per-neighbor attribute.  This patch allows it to appear directly in the 'neighbor' scope.

I think this was probably the original intention since (almost) every other attribute is the same between the [group scope](https://github.com/Exa-Networks/exabgp/blob/3.4/lib/exabgp/configuration/ancient.py#L1277) and the [neighbor scope](https://github.com/Exa-Networks/exabgp/blob/3.4/lib/exabgp/configuration/ancient.py#L1518).

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/504)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,7 +6,9 @@ Version explained:
 Version 3.4.18
  * Fix: Configuration parser does not accept configs without neighbors.
     patch by doddt
-    
+ * Fix: 'connect' keyword is now also allowed in neighbor scope
+    patch by: Stacey Sheldon (Corsa)
+
 Version 3.4.17
  * Fix: does not accept IPv6 as router-id
     reported by: yuriya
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Thank you for finding this issue - an bad oversight - hopefully there is not too many of these 😜 

As we never had a 4.0.0 release, this is should not be listed as fix: the whole configuration code is new, if we were to indicate all the fixes of breakage in the configuration done during the while 4.0 cycle the CHANGELOG would become useless.

Forward port of issues fixed in 3.4 are not listed neither. I will copy all the 3.4.x release before releasing 4.0 - which will include these fixes.

I hope it makes sense. So after asking you to add a CHANGELOG, I am now doing the opposite .. Sorry to be such a pain .. Hopefully it makes sense.

solidgoldbomb:
Sure, no problem.  Are you OK with a force push on this branch to remove the changelog from the commit?

thomas-mangin:
Sorry - this patch was for 3.4 - all good 😄

PR #506, opened 2016-10-01
Fix ipv4 unicast eor json encoding

solidgoldbomb:
The IPv4 Unicast EOR messages were being serialized into invalid JSON across the API.  They were also being incorrectly serialized as an NLRI within an announce which differs from the encoding of EOR in all other AFIs/SAFIs so this embedding seems unintentional.

This patch set fixes the structure of the IPv4 unicast EOR to match the other families.  It also ensures that the string encoding of these EORs is provided over the API.  AFI/SAFI was being encoded as ints.

Old IPv4 EOR (invalid json, ints for AFI/SAFI)
```
...
		"message": {
			"update": {
				"announce": {
					"1 1": {
						"null": ["eor": {
							"afi": "1",
							"safi": "1"
						}]
					}
				}
			}
		}
...
```

New IPv4 EOR (valid json, strings for AFI/SAFI)
```
...
        "message": {
            "eor": {
                "afi": "ipv4", 
                "safi": "unicast"
            }
        }
...
```

IPv6 EOR for reference (unchanged in this patch set)
```
...
        "message": {
            "eor": {
                "afi": "ipv6", 
                "safi": "unicast"
            }
        }
...
```

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/506)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -153,6 +153,8 @@ Version 4.0.0
    - encoding is now valid JSON
    - dropped auto-generated name for flowspec updates
    - format for flowspec updates is now a list of dicts where each dict contains a single flowspec rule
+ * Fix: API encoding of IPv4 Unicast EOR messages were being encoded as NLRI updates
+   patch by: Stacey Sheldon (Corsa)
 
 Version 3.4.10
  * Fix: Fix parsing attributes with PARTIAL flag set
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Thank you.

PR #518, opened 2016-10-19
3.4

borjam:
Fixed ttl-security for IPv6.

The previous fix did only work on IPv4. On IPv6 it caused an exception because the IP_TTL socket option is only valid for IPv4, while on IPv6 UNICAST_HOPS must be used.

Added a function for setting TTL on IPv6 and an address family check to choose the right function.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/518)


thomas-mangin:
Any chance to see a forward port for master please ?

thomas-mangin:
See - https://github.com/Exa-Networks/exabgp/issues/519

borjam:
Sure!

PR #520, opened 2016-10-20
Added ttl-security (now called outgoing-ttl) feature to IPv6.

borjam:
Support for outgoing-ttl (previously known as ttl-security).

Should we add some warning for the deprecated "ttl-security" option? Or maybe it would be a good idea to accept ttl-security as an equivalent to outgoing-ttl *and* log a warning about a deprecated option?

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/520)


thomas-mangin:
Thank you very much :-)

PR #526, opened 2016-11-03
[WIP] Healthcheck support for specifying IP addresses as networks

benagricola:
This implements some new functionality if the python 3 compatible ipaddress module is available (`py2-ipaddress`).

## Without `py2-ipaddress`
Passing /32 IPs only:

```
# ./healthcheck.py --ip=1.2.3.4 --ip=1.2.3.5 --ip=1.2.3.6 --no-ip-setup
INFO[healthcheck] send announces for UP state to ExaBGP
announce route 1.2.3.4/32 next-hop self med 100
announce route 1.2.3.5/32 next-hop self med 101
announce route 1.2.3.6/32 next-hop self med 102
```

Passing CIDR notation networks only:
```
# ./healthcheck.py --ip=1.2.3.0/30 --ip=1.2.3.4/30 --no-ip-setup
usage: healthcheck.py [-h] [--debug | --silent | --syslog-facility [FACILITY]
                      | --no-syslog] [--name NAME] [--config FILE]
                      [--pid FILE] [--user USER] [--group GROUP]
                      [--interval N] [--fast-interval N] [--timeout N]
                      [--rise N] [--fall N] [--disable FILE] [--command CMD]
                      [--next-hop IP] [--ip IP] [--no-ip-setup]
                      [--label LABEL] [--start-ip N] [--up-metric M]
                      [--down-metric M] [--disabled-metric M] [--increase M]
                      [--community COMMUNITY]
                      [--large-community LARGECOMMUNITY] [--as-path ASPATH]
                      [--withdraw-on-down] [--execute CMD] [--up-execute CMD]
                      [--down-execute CMD] [--disabled-execute CMD]
healthcheck.py: error: argument --ip: invalid IPAddress value: '1.2.3.0/30'
```
 _(note: same behaviour occurs with a mix of CIDR notation and IPs)_

## With `py2-ipaddress`
Passing /32 IPs only:

```
./healthcheck.py --ip=1.2.3.4 --ip=1.2.3.5 --ip=1.2.3.6 --no-ip-setup
INFO[healthcheck] send announces for UP state to ExaBGP
announce route 1.2.3.4/32 next-hop self med 100
announce route 1.2.3.5/32 next-hop self med 101
announce route 1.2.3.6/32 next-hop self med 102
```

Passing CIDR notation networks only:
```
# ./healthcheck.py --ip=1.2.3.0/30 --ip=1.2.3.4/30 --no-ip-setup
INFO[healthcheck] send announces for UP state to ExaBGP
[...]

PR #291, opened 2015-06-09
Add initial specfile redhat based distros

abn:
This was developed as part of packaging efforts for the [CloudRouter Project](https://cloudrouter.org/) and is pending merge [here](https://github.com/cloudrouter/cloudrouter/pull/84).

Hope this is a useful addition for  the exabgp project.

[Review on Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/291)

PR #528, opened 2016-11-03
Fix loopback IP lookups for non /32 IPs

benagricola:
Related to the changes in #526 - I found an issue with the `healthcheck.py` IP adding system that assumes that all loopback IPs it scans are /32s (line #291), instantiating `ip_address` with just the IP address and not the parsed prefix length.

This causes an issue when specifying a network using --ip - the network itself is assigned to the loopback device before the addresses are deaggregated (which IMO is nice since it keeps the applied IP list short), but the next time healthcheck is run it cannot find the already existing network and throws an error trying to add it.

This patch fixes the issue in the short term by passing 'netmask' retrieved from the IP command into `ip_address` along with the address itself.

_note: I suspect this will be an issue on non-linux systems using `ipconfig lo0`, but I have no way to test - it looks  like the netmask field parsed out from the ifconfig regex is in a different format, the actual netmask as a hex number. I suspect the correct field in this case is the `mask` field but I can't be sure. Maybe it makes sense to make sure that the prefix length field is called the same in both regex so we can use it in this manner to identify non-/32 ranges?_

edit: Amended commit fixes the above by renaming `netmask` field to `mask` field in the 'linux' branch of `loopback_ips`.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/528)


thomas-mangin:
@vincentbernat LGTM .. just merge if you are happy too.

PR #534, opened 2016-11-07
Fix pidfile readline issue and invalid fd on nonrunning process

benagricola:
Hi @thomas-mangin 

This fixes 2 issues with the new pidfile code where:

1) Pidfile exists and we attempt to `.readline()` on an already-read string
2) Pidfile exists but `check_pid` returns false (process not running i.e. stale) - we do nothing, and later code expects `fd` to be set to the pidfile but it isn't.

Quick tests indicate this works effectively to detect stale pidfiles.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/534)

### DIFF
--- a/lib/exabgp/reactor/daemon.py
+++ b/lib/exabgp/reactor/daemon.py
@@ -65,10 +65,13 @@ class Daemon (object):
 			fd = os.open(self.pid,flags,mode)
 		except OSError:
 			try:
-				pid = open(self.pid,'r').read().readline()
+				pid = open(self.pid,'r').readline()
 				if self.check_pid(int(pid)):
 					self.logger.daemon("PIDfile already exists and program still running %s" % self.pid)
 					return False
+				else:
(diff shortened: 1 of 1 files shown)


thomas-mangin:
sorry it should have been readline().strip() .. will fix.

thomas-mangin:
Fixed. Thank you for the report of the issue 👍

benagricola:
Hi Thomas,

There was another fix in this commit (line 74 in the changes) which stops an error occurring when the pidfile exists but there's no matching process running - I don't think this made it into your fixed commit :)

thomas-mangin:
Ok - we dropped privileges will add the second part.

thomas-mangin:
About the security implication about
`fd = os.open(self.pid,flags ^ os.O_EXCL,mode)`
as we already drop privileges it should be fine, I can not see how a symlink timing attack could work ...

PR #535, opened 2016-11-07
When group-updates is enabled, send one UPDATE per address family

brijohn:
Currently exabgp is disabling group-updates when sending non ipv4/unicast families to prevent more then one MP_REACH/MP_UNREACH attribute from appearing in the UPDATE message. 

With this fix instead of creating a separate UPDATE message for each NLRI when we have group-updates enabled, we will instead create a single UPDATE for each family in our list of NLRIs.  This should still prevent duplicate path attributes, but also allow  grouping of announcements and withdrawals were possible.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/535)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,8 @@ Version 3.4.18
     reported by: Ben Agricola
  * Fix: Update RIB cache families on configuration reload
     patch by: Brian Johnson
+ * Fix: When group-updates is enabled, send one UPDATE per address family
+    patch by: Brian Johnson
 
 Version 3.4.17
  * Fix: does not accept IPv6 as router-id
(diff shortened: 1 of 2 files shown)


thomas-mangin:
well spotted ! Could I ask you for the same fix for master too please ?

The patch for master does not need a CHANGELOG entry but the patch for 3.4 should include an 'Change' entry defining the behaviour change and giving you credit.

thomas-mangin:
'Fix' in changelog define bug fix. Technically the current behavior is not buggy (the current code is "just" sub-optimal), however it does CHANGE the how Update are generated, so a "Change" line would be better as it would warn users that it may affect them. Sorry for being so pedantic 😜

brijohn:
You went and merged before I could update the Fix to Change :smile:

thomas-mangin:
Yes done it 👅

PR #539, opened 2016-11-10
Read and discard command confirmation from exabgp

benagricola:
Now that ExaBGP responds to commands with `done\n`, it's possible for ExaBGP + healthcheck.py to get stuck unless healthcheck.py reads the response to announces / withdrawls from ExaBGP.

This simply reads the contents of stdin on the healthcheck process after sending commands to ExaBGP to make sure ExaBGP doesn't get stuck trying to write output to healthcheck which it isn't accepting.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/539)

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -455,8 +455,10 @@ def loop(options):
                         options.as_path)
             logger.debug("exabgp: %s %s", command, announce)
             print("{0} {1}".format(command, announce))
+            # Flush command and wait for confirmation from ExaBGP
+            sys.stdout.flush()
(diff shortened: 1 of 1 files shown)


benagricola:
Note: moved `sys.stdout.flush()` into the command loop since the easiest way to implement this is to wait for 'done' line for each submitted command. I believe it is pretty much necessary to send, flush and readline in that order for each command since otherwise there's a possibility ExaBGP runs out of stdin buffer to the healthcheck process while writing, and the healthcheck process is stuck attempting to write to stdout which ExaBGP is not attempting to read at that point.

PR #546, opened 2016-11-29
Update setup.py

aabdnn:
Installing exabgp from source, or building a tarball or wheel requires setuptools. However, setuptools is NOT required when installing a pre-built wheel of exabgp on a target system. This small patch corrects this mistake in the setup.py script.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/546)


thomas-mangin:
Thank you for this fix Anand.

PR #551, opened 2016-12-10
Adds parsing support for RFC 7752

ruissalo:
This commit adds BGP Link State (RFC 7752) parsing support for both the MP NLRI and the link state attribute. It doesn't support Link State announcement.
Implementation was tested with Junos 15.1F using the json parser. 
While announcement is not really implemented, I think it can be useful as it is for users looking to monitor the IGP topology and traffic engineering attributes.  Class registering was mostly cargo-culted from existing NLRI classes. I can add a wiki page too if you think it's a good idea.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/551)

### DIFF
--- a/doc/man/exabgp.1
+++ b/doc/man/exabgp.1
@@ -390,5 +390,7 @@ Capabilities Advertisement with BGP-4
 Autonomous-System-Wide Unique BGP Identifier for BGP-4
 .It RFC 6608
 Subcodes for BGP Finate State Machine Error
+.It RFC 7752
+North-Bound Distribution of Link-State and Traffic Engineering (TE) Information Using BGP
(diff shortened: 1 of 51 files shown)


thomas-mangin:
Hello,

Sorry for the time it took me to reply - I am sure I will want to rename things here and there to make my OCD happy :-) but otherwise it looks fine.

Thank you very much,

Thomas

ruissalo:
hi Thomas, np, I'll try to help with bugs reports the best I can. Thanks for merging the PR :)

thomas-mangin:
I still need to fully review the code I may need to ask you thing when I do :)

PR #552, opened 2016-12-12
Fixed invalid FSM constant in state table

arthurec:
Noticed a typo in exabgp/lib/reactor/peer.py:detailed_link_status

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/552)


thomas-mangin:
The FSM state is sooo no not right, it needs a full rewrite but this is indeed a bad typo :-)
Thank you.

PR #558, opened 2017-01-03
Don't always add ipv4/unicast family

brijohn:
Currently ExaBGP seems to be adding the IPv4/unicast family even when not requested to the list of supported families.  There does not seem to be a good reason for this and it can cause unexpected results, so this removed the lines of code responsible.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/558)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -154,6 +154,8 @@ Version 4.0.0
    patch by: Stacey Sheldon (Corsa)
  * Fix: Update RIB cache families on configuration reload
     patch by: Brian Johnson
+ * Fix: Do not add IPv4/unicast family unless specifically configured
+    patch by: Brian Johnson
 
 Version 3.4.10
  * Fix: Fix parsing attributes with PARTIAL flag set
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Hi @brijohn,

Thank you very much for the patch.

Could you please tell me why from your point of view it caused a surprise. You can always announce IPv4 unicast routes on a BGP session ...

I think I tried to remove this some time ago and got some complain when I did. I will need to look at the rationale for the current behaviour and see if indeed it should be changed.

Thanks.

Thomas

brijohn:
It causes a surprise based on the fact that we can now negotiate a supported family that we are not expecting based on our configuration. This can in our usage also end up generating an alert due to the unexpected family. 

Besides that, this just seems wrong to add a family I didn't add to the list of supported families. As far as I can tell there is nothing in the RFCs that state a BGP session should always support IPv4/unicast.  If only want to support say ipv6/unicast or ipv4/flowspec i should be able to without having ipv4/unicast tossed in as well. Also both juniper and cisco routers that I am familiar with are perfectly fine with not advertising support for ipv4/unicast.

thomas-mangin:
will make a mental note for when I get the next complain ... :-)

PR #559, opened 2017-01-03
Fix api configuration parsing and off by one error

brijohn:
While testing the API options for neighbors i came across two issues preventing the sending of messages to sub processes.

The first one was that  the ParseAPI class was using an instance version of the _built dict during the parse phase, but when building the neighbor it was using the class version of the dict which was still empty.  I changed the parse phase so it should always use the class version as well.

The second was a simple off by one error when a message was being sent that was not grabbing the right offset for the message type.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/559)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -160,6 +160,10 @@ Version 4.0.0
     patch by: Brian Johnson
  * Fix: Do not add IPv4/unicast family unless specifically configured
     patch by: Brian Johnson
+ * Fix: ParseAPI should always use the class version of the _built dict
+    patch by: Brian Johnson
+ * Fix: Off by one error when getting message type for send-* api
+    patch by: Brian Johnson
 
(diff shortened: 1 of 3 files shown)


thomas-mangin:
whoa . nice patches ! Thanks a lot.

thomas-mangin:
Could you please tell me why you needed to use type(self) .. I am not sure why this was required.

brijohn:
I used the type(self) to ensure that the class version of _built was being used. Without that when the classmethod extract was being called the _built dictionary was empty, since when constructing the _built dictionary python was creating and using an instance version of the dictionary. The use of type(self) is just to ensure that the class version is always used.

thomas-mangin:
Thank you. Understood, I will look why the variable was shadowed as it should not have been.

PR #563, opened 2017-01-05
The md5-ip config options should be an IP address not a RouterID

brijohn:
When this is defined as a router_id a parse error will be generated if an IPv6 address is specified.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/563)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -164,6 +164,8 @@ Version 4.0.0
     patch by: Brian Johnson
  * Fix: Off by one error when getting message type for send-* api
     patch by: Brian Johnson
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Hello, I have merged but fix of bugs for code which does not exist in 3.4 is not something to add to the change log :-)

brijohn:
Oh, I assumed a Fix status in the changelog was for any bugfix.

thomas-mangin:
Only If the bug did exist in the previous release - it is to inform of the change between version.

PR #566, opened 2017-01-06
Fix Update.messages so that it won't generate invalid UPDATEs

brijohn:
This is revisiting the issue with sending an UPDATE message that contains multiple MP_REACH or MP_UNREACH attributes.

During some of our testing we encountered another way in which the current code allows this to happen. If you send enough nlris before sending a route flush (in our case we got this to happen using auto flush while handling a route refresh), the current code will determine that all of the nlris can't fit inside a single UPDATE message and end generating multiple mpreach attributes with a single nlri. These then get packed into one or more UPDATE messages which is invalid.

This code rewrites how UPDATE messages are generated so that it never sticks more then one MP_(UN)REACH attribute in a single UPDATE.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/566)

### DIFF
--- a/lib/exabgp/bgp/message/update/__init__.py
+++ b/lib/exabgp/bgp/message/update/__init__.py
@@ -104,147 +104,72 @@ class Update (Message):
 	def messages (self, negotiated, include_withdraw=True):
 		# sort the nlris
 
-		add_nlri = []
-		del_nlri = []
-		add_mp = {}
-		del_mp = {}
+		nlris = []
+		mp_nlris = {}
 
 		for nlri in self.nlris:
 			if nlri.family() in negotiated.families:
 				if nlri.afi == AFI.ipv4 and nlri.safi in [SAFI.unicast, SAFI.multicast]:
-					if nlri.action == OUT.ANNOUNCE:
-						add_nlri.append(nlri)
-					elif include_withdraw:
-						del_nlri.append(nlri)
+					nlris.append(nlri)
 				else:
-					if nlri.action == OUT.ANNOUNCE:
-						add_mp.setdefault(nlri.family(),[]).append(nlri)
-					elif include_withdraw:
-						del_mp.setdefault(nlri.family(),[]).append(nlri)
+					mp_nlris.setdefault(nlri.family(), {}).setdefault(nlri.action, []).append(nlri)
 
-		if not add_nlri and not del_nlri and not add_mp and not del_mp:
+		if not nlris and not mp_nlris:
 			return
 
-		if add_nlri or add_mp:
-			attr = self.attributes.pack(negotiated,True)
+		attr = self.attributes.pack(negotiated, True)
+
+		# Withdraws/NLRIS (IPv4 unicast and multicast)
+		msg_size = negotiated.msg_size - 19 - 2 - 2 - len(attr) # 2 bytes for each of the two prefix() header
+		withdraws = ''
+		announced = ''
+		for nlri in nlris:
+			packed = nlri.pack(negotiated)
+			if len(announced + withdraws + packed) > msg_size:
+				if not withdraws and not announced:
+					raise Notify(6,0,'attributes size is so large we can not even pack one NLRI')
+				yield self._message(Update.prefix(withdraws) + Update.prefix(attr) + announced)
+				if nlri.action == OUT.ANNOUNCE:
+					announced = packed
+					withdraws = ''
+				elif include_withdraw:
+					withdraws = packed
+					announced = ''
+			else:
+				if nlri.action == OUT.ANNOUNCE:
+					announced += packed
+				elif include_withdraw:
+					withdraws += packed
+
+		if mp_nlris:
+			for family in mp_nlris.keys():
+				afi, safi = family
+				mp_reach = ''
+				mp_unreach = ''
+				mp_announce = MPRNLRI(afi, safi, mp_nlris[family].get(OUT.ANNOUNCE, []))
+				mp_withdraw = MPURNLRI(afi, safi, mp_nlris[family].get(OUT.WITHDRAW, []))
+
+				for mprnlri in mp_announce.packed_attributes(negotiated, msg_size - len(withdraws + announced)):
+					if mp_reach:
+						yield self._message(Update.prefix(withdraws) + Update.prefix(attr + mp_reach) + announced)
+						announced = ''
+						withdraws = ''
+					mp_reach = mprnlri
+
+				if include_withdraw:
+					for mpurnlri in mp_withdraw.packed_attributes(negotiated, msg_size - len(withdraws + announced + mp_reach)):
+						if mp_unreach:
+							yield self._message(Update.prefix(withdraws) + Update.prefix(attr + mp_unreach + mp_reach) + announced)
+							mp_reach = ''
+							announced = ''
+							withdraws = ''
+						mp_unreach = mpurnlri
+
+				yield self._message(Update.prefix(withdraws) + Update.prefix(attr + mp_unreach + mp_reach) + announced) # yield mpr/mpur per family
+				withdraws = ''
+				announced = ''
 		else:
-			attr = ''
-
-		# withdrawn IPv4
-
-		packed_del = ''
-		msg_size = negotiated.msg_size - 19 - 2 - 2  # 2 bytes for each of the two prefix() header
-
-		while del_nlri:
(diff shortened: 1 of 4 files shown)


thomas-mangin:
Hello Brian,

Thank you for the effort and work here. I have taken the rest of the day (16:30 to 18:00 GMT) to work on ExaBGP.

This is a large rewrite of quite an important part of the code generation. It will take some time to review. Any way to interactively (call, video call, IM) review this patch to understand why you done what you did please ?

Thomas

brijohn:
Thomas,
I have a around 30 minutes available at around 16:30 GMT, but after that I will be busy till after sometime after 18:00 GMT. If you want to setup a time to talk about these changes can you give me an email and we can setup a time.

lib/exabgp/bgp/message/update/__init__.py
```diff
@@ -104,147 +104,72 @@ def split (data):
...
+				mp_withdraw = MPURNLRI(afi, safi, mp_nlris[family].get(OUT.WITHDRAW, []))
+
+				for mprnlri in mp_announce.packed_attributes(negotiated, msg_size - len(withdraws + announced)):
+					if mp_reach:
+						yield self._message(Update.prefix(withdraws) + Update.prefix(attr + mp_reach) + announced)
+						announced = ''
+						withdraws = ''
+					mp_reach = mprnlri
+
+				if include_withdraw:
```
thomas-mangin: I think the withdraw should be sent before the annouce, as otherwise you can remove an update for which you have a new announcement ... I will fix change the patch once merged.

lib/exabgp/bgp/message/update/attribute/mpurnlri.py
```diff
@@ -41,29 +41,26 @@ def __eq__ (self, other):
...
-
-		for (pafi,psafi),nlris in mpurnlri.iteritems():
-			payload = pafi + psafi + ''.join(nlris)
+			if nlri.family() != self.family(): # nlri is not part of specified family
+				continue
+			mpurnlri.append(nlri.pack(negotiated))
 
-			if self._len(payload) <= maximum:
+		payload = ''.join([self.afi.pack(), self.safi.pack()])
+		for nlri in mpurnlri:
```
thomas-mangin: We should check that the prefix fit in maximum but I do not believe it can happen ..

lib/exabgp/bgp/message/update/attribute/mprnlri.py
```diff
@@ -69,29 +69,19 @@ def packed_attributes (self, negotiated):
...
-				for nlri in nlris:
-					yield self._attribute(
-						pafi + psafi +
-						chr(len(nexthop)) + nexthop +
-						chr(0) + nlri
-					)
+			yield self._attribute(payload)
 
 	def pack (self, negotiated):
 		return ''.join(self.packed_attributes(negotiated))
```
thomas-mangin: packed_attributes is not passed maximum and maximum has no default AFAICS

lib/exabgp/bgp/message/update/attribute/mpurnlri.py
```diff
@@ -41,29 +41,26 @@ def __eq__ (self, other):
...
-			# This will not generate an optimum update size..
-			# we should feedback the maximum on each iteration
-
-			for nlri in nlris:
-				yield self._attribute(pafi + psafi + nlri)
+			payload = ''.join([payload, nlri])
+		yield self._attribute(payload)
 
 	def pack (self, negotiated):
 		return ''.join(self.packed_attributes(negotiated))
```
thomas-mangin: no maximum for packed_attributes here neither.

lib/exabgp/bgp/message/update/__init__.py
job: if you send a withdraw before the announce, wouldn't you have a very brief outage? if you are announcing a new path for an NLRI, do you even need to send a withdraw?

brijohn: This case seems a little odd to me, and I'm not sure if the BGP RFCs say anything about it. 

But wouldn't the reverse be true as well, If you withdraw first and then announce you get end up with an announced route that you had meant to withdraw, correct? Again I'm not clear what the expected behavior in this case is.

thomas-mangin: In the other case, you update your NLRI to then remove it .. which is surely a much larger outage 😜 The code was previously grouping the remove and adding in the same UPDATE AFAICR, not sure what the new does ..

thomas-mangin: If @brijohn tells me he considered the scenario and is happy - good with me - just trying to make sure it is not overlooked.

lib/exabgp/bgp/message/update/attribute/mprnlri.py
brijohn: Easy to fix. I do have a question though do you know if the pack function on mprlnri is called anywhere? I couldn't actually find anyplace in the code that was using it myself. The problem with this pack function is that if it is ever called and used anywhere it can still generate multiple nlris. If its not used I think the best option would be to remove it.

lib/exabgp/bgp/message/update/attribute/mpurnlri.py
brijohn: See as above.

lib/exabgp/bgp/message/update/__init__.py
thomas-mangin: btw, the way would be to not have the withdraw if we have the announce, as a announce on a announce is a replacement - as I wrote this code years ago - I am not sure myself what I did back then .. And did not write enough test too :-(

lib/exabgp/bgp/message/update/attribute/mprnlri.py
thomas-mangin: Let me use atom for a global search ..
It is only used in update/__init__.py, attributes/mpu?rnlri.py.

lib/exabgp/bgp/message/update/__init__.py
job: If the path changes, you just need to send an UPDATE with the new path attributes. If the path really is gone, then yes, withdraw. On the receiving side the new UPDATE just overwrites the previous known path.

lib/exabgp/bgp/message/update/attribute/mprnlri.py
thomas-mangin: Sorry @brijohn - I may have mis-understood your question. This code should only be used within the class Update AFAICR - but I may miss a case - the issue with dynamic language :-( ..

lib/exabgp/bgp/message/update/__init__.py
thomas-mangin: Having looked and thought. The initial dict will only contain a withdraw and/or announce (as the check is done in the RIB code) so the order should not matter .. a long thread for nothing it seems 🤦‍♂️

lib/exabgp/bgp/message/update/attribute/mprnlri.py
thomas-mangin: Attributes classes and Update it is.

lib/exabgp/bgp/message/update/__init__.py
brijohn: So Yeah I was thinking about this, and came to the same conclusion i think Thomas just mentioned. At this point in the code we should never have an announce and withdraw for the same route going out. That checking should be done i think in the RIB code, so by the time we get here it shouldn't as far as i can tell matter if we do announces or withdraws first.

lib/exabgp/bgp/message/update/attribute/mprnlri.py
brijohn: So since it was a bug I added a default value for maximum into the code. I'm still not sure that the pack method is every called on these two classes though since I never saw any exceptions or errors during testing that I would have expected had that function ever been called.

lib/exabgp/bgp/message/update/attribute/mpurnlri.py
brijohn: Yeah not sure this can ever happen either, but added a few extra checks in the code to raise a Notify exception if it does.

thomas-mangin:
@brijohn ready for merge ?

PR #567, opened 2017-01-06
Add API for show routes by type (static/flow/l2pn)

brijohn:
This adds a feature we find useful that will allow you to show routes by type (static, flow or l2vpn routes.).  In addition when using this new API it will only show routes that are being advertised (those belonging to a negotiated family) for each neighbor.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/567)


lib/exabgp/rib/store.py
```diff
@@ -54,7 +54,7 @@ def clear (self):
 
 	def sent_changes (self, families=None):
 		# families can be None or []
-		requested_families = self.families if not families else set(families).intersection(self.families)
+		requested_families = self.families if families == None else set(families).intersection(self.families)
```
thomas-mangin: should be "families is None" if you do not mind changing it .. Otherwise Happy to merge.

thomas-mangin:
@brijohn You have a mail to make working on the code easier. Happy to continue working as we do if you want me to review the code but I am happy with your past contributions. I get notifications of changes anyway and will review automatically (and tweak if required). We can work by exception.

PR #569, opened 2017-01-22
not to terminate never created processes

dontreboot:
When USR2 is sent after configuration file which has been updated with additional process(es) causes exabgp to cause because _terminate() to throw a 'KeyError' exception

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/569)


dontreboot:
This should fix the bug #570 I reported

thomas-mangin:
Thank you

PR #571, opened 2017-01-31
[v3.4] allow sudo to setup ips and manage loopback dynamically

ahmet2mir:
# 2 features in this PR

To install and test it
```
pip install git+https://github.com/ahmet2mir/exabgp.git@sudo_dyn_ip_34
```
## 1 - Use sudo instead of drop privileges

When I take a look to apache2 or nginx configuration, both of them launch a daemon as root, then this deamon launch child with the specified service user.

httpd example

```
root     15174       1   /usr/sbin/httpd
apache   11502   15174   /usr/sbin/httpd
```

But running exabgp as root then remove privileges could be very dangerous (I remember some hacking in C bypassing the setuid, very old in my memory :)) So I add an option to prepend cmd ip with sudo.

Current `lo` conf

```
[root@myhost  ~]# ip a s dev lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
```

Start the daemon, run as exabgp

```
[root@myhost  ~]# systemctl start exabgp
[root@myhost  ~]# ps -ef | grep exabgp
exabgp   14259     1  0 10:15 ?        00:00:00 /bin/python -m exabgp healthcheck --withdraw-on-down --next-hop 10.1.0.1 --disable /etc/exabgp/maintenance/myhost.lan --cmd /bin/curl -sf http://localhost:8080 -o /dev/null --name myvip.lan --up-metric 100 --ip 10.2.0.1
```

Get the logs

```
[root@myhost  ~]# journalctl -fxu exabgp
Jan 31 10:16:39 rancher-slave02.sys.lab.ingenico.com healthcheck-myvip.lan[15089]: Uncaught exception: Command '[u'ip', u'address', u'add', '10.2.0.1/32', u'dev', u'lo']' returned non-zero exit status 2
Traceback (most recent call last):
File "/usr/lib/python2.7/site-packages/exabgp/application/healthcheck.py", line 598, in main
  setup_ips(options.ips, options.label, options.sudo)
File "/usr/lib/python2.7/site-packages/exabgp/application/healthcheck.py", line 329, in setup_ips
  cmd, stdout=fnull, stderr=fnull)
File "/usr/lib64/python2.7/subprocess.py", line 542, in check_call
  raise CalledProcessError(retcode, cmd)
[...]

PR #572, opened 2017-01-31
[v4] allow sudo to setup ips and manage loopback dynamically

ahmet2mir:
Same PR #571 but for v4
Only  `toremove = existing.intersection(ips)`  is different (using remove_ip from master)

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/572)

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -106,6 +106,8 @@ def parse():
                    default="daemon",
                    help=("log to syslog using FACILITY, "
                          "default FACILITY is daemon"))
+    g.add_argument("--sudo", action="store_true", default=False,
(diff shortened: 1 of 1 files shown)


vincentbernat:
There is some noise in the diff.

tmmorin:
one comment: oslo_rootwrap is a nice helper to use sudo, add per command fine-grained filters for more thorough checks, etc.

https://pypi.python.org/pypi/oslo.rootwrap

ahmet2mir:
@vincentbernat oups, get-pip and diff stuff, removed them.
@tmmorin as you can see in requirements, there is only one external package (ipaddr), so IMHO, I think that exabgp team wan't to be 'free of deps'.

PR #574, opened 2017-02-09
Backport for jessie-backports

gaudenz:
There is also a signed tag corresponding to the source I uploaded to jessie-backports in my repository. Don't know how you can import this into the upstream repository with the github web interface, but it's certainly doable with the git command.

This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/574)


vincentbernat:
Thanks! I'll merge that offline to be sure to get tags and all.

PR #579, opened 2017-02-27
Include the IP prefix length in JSON output and fix error message output

jsynack:
The JSON output for IP Prefix Advertisements (type 5) now includes the prefix length (integer)

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/579)


thomas-mangin:
Thank you

PR #583, opened 2017-03-08
batch of python3 compatibility fixes

tmmorin:
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/583)


thomas-mangin:
but ord_ and chr_ which I find horrible as name - very nice.

PR #585, opened 2017-03-14
Misc filename header fixes and typos

jsynack:
Updated filename headers to be the proper filename.py as encountered.

Updated a few error strings / text messages / comments.

OPERATIONAL NOTE: Corrected typo ('toplogy' -> 'topology') in BGPLS JSON output -
this may break anyone that used the previous token (the typo) in the
JSON.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/585)

### DIFF
--- a/lib/exabgp/__main__.py
+++ b/lib/exabgp/__main__.py
@@ -1,6 +1,6 @@
 # encoding: utf-8
 """
-__init__.py
+__main__.py
 
 Created by Thomas Mangin on 2010-01-15.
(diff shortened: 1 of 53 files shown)


thomas-mangin:
Thank you very much, much appreciated.

thomas-mangin:
The BGP-LS code was only merged recently at signing @ruissalo here so he can update his code if he has the same typo in other places.

PR #591, opened 2017-03-27
Added support for draft-ietf-idr-flowspec-interfaceset

sli35:
- added a scope configuration stanza for flow routes
to configure interface-set communities
- both transitive and non transitive instance are supported

Example of configuration:
```
flow {
        route MYROUTE {
                match {
                        source 1.1.1.1/32;
                        destination 2.2.2.2/32;
                }
                scope {
                        interface-set [  transitive:input:3:3 non-transitive:output:12:1 transitive:input-output:1:1 ];
                }
                then {
                        accept;
                }
        }
}
```

Signed-off-by: Stephane Litkowski <stephane.litkowski@orange.com>

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/591)


thomas-mangin:
Hello Stephane,

Overall the patch is fine, but before merging, I would really appreciate if you could do the following (as otherwise I will need to do it myself once merged):
 - convert of the space used for indentation as tabs (in line with the rest of the cod)
 - `InterfaceSet` direction is a string but all the other contructor for the class seems to be raw values, it should be a boolean
 - `InterfaceSet` the unpack function does not even pass the expected string but `InterfaceSet(0,` it should be `InterfaceSet(False,`
 - InterfaceSet use a dict with get(value,'default) for the `__repr__`
 - `def _interface_set (data):` the test should be inverted at the start (`if data.count(':') == 3: raise`) so the whole code does not need to be indented (and updated to the new `InterfaceSet` API)

Otherwise looks good and congratulation on understanding the configuration parsing code 👍

thomas-mangin:
Also it says that this branch is conflicting so could you please update your version of master and re-submit against HEAD - thank you.

sli35:
Hi Thomas,


Thanks for your feedback

Sorry for the tabs, I’m using VI to edit the code and it does align correctly the tab with the existing code. I will fix it with another editor.
I’m not sure to understand correctly the two following comments:
-          InterfaceSet direction is a string but all the other contructor for the class seems to be raw values, it should be a Boolean
The direction cannot be a Boolean as we need 3 values
-          InterfaceSet use a dict with get(value,'default) for the __repr__
Could you explain more what you are expecting ?

Thanks

thomas-mangin:
The function `_interface_set` should compare the string with 'transitive', you moved some of the logic of the parser in the class:

The class `InterfaceSet` has this `def __init__ (self, trans` below trans is compared with the string `'transitive'`. It is then stored as the boolean result of the comparaison.

Your own code does not use a string but the value `0` when calling the `__init__` and this `trans` should clearly be a boolean. 0 in that case evaluate to False (as it is not the word `'transitive'`) but is pure luck. It should have been passed as `False`.

For the dict trick:

```
class InterfaceSet ...:
   names = {
      1: 'input',
      2: 'output',
      3: 'input-output'
   }

  def __repr__ (self):
     str_direction = self.names(self.direction,str(self.direction))
```

I hope this clarify what I meant :-)

sli35:
I normally fixed the issues. Let me know if I missed something.

lib/exabgp/bgp/message/update/attribute/community/extended/flowspec_scope.py
```diff
@@ -42,20 +45,13 @@ def __init__ (self, trans, asn, target, direction, community=None):
...
 	def __repr__ (self):
-		if (self.direction == 1 ):
-			str_direction = 'input'
-		elif (self.direction == 2 ):
-			str_direction = 'output'	
-		elif (self.direction == 3):
-			str_direction = 'input-output'
-		else:
-			str_direction = str(self.direction)	
+		str_direction = self.names[self.direction]
```
thomas-mangin: very sorry, my example was wrong, it should read:
`str_direction = self.names.get(self.direction,str(self.direction))`

sli35:
Fixed

sli35:
Sorry there is something bad that happened with the rebase. Trying to fix it.

PR #597, opened 2017-04-05
Allow local_address to be optional

brijohn:
In one of our use cases we would like to be able to set up a peer without specifying the local ip address to use for that connection.

This patch allows the local_address to be left out in which case the correct local IP will be determined by the OS using its current routing table.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/597)

### DIFF
--- a/lib/exabgp/reactor/loop.py
+++ b/lib/exabgp/reactor/loop.py
@@ -229,7 +229,7 @@ class Reactor (object):
 							# * later if it should be called again but has no work atm
 							# * close if it is finished and is closing down, or restarting
 							if action == ACTION.CLOSE:
-								self.unschedule(peer)
+								self.unschedule(key)
 								peers.discard(key)
 							# we are loosing this peer, not point to schedule more process work
 							elif action == ACTION.LATER:
@@ -252,7 +252,7 @@ class Reactor (object):
 								peer = self.peers[key]
 								neighbor = peer.neighbor
 								# XXX: FIXME: Inet can only be compared to Inet
-								if connection.local == str(neighbor.peer_address) and connection.peer == str(neighbor.local_address):
+								if connection.local == str(neighbor.peer_address) and (not neighbor.local_address or connection.peer == str(neighbor.local_address)):
 									if peer.incoming(connection):
 										found = True
 										break
@@ -440,9 +440,8 @@ class Reactor (object):
 		self.processes.start()
 
 	def unschedule (self, peer):
-		key = peer.neighbor.name()
-		if key in self.peers:
-			del self.peers[key]
+		if peer in self.peers:
+			del self.peers[peer]
 
 	def answer (self, service, string):
 		self.processes.write(service,string)
--- a/lib/exabgp/reactor/loop.py
+++ b/lib/exabgp/reactor/loop.py
@@ -252,7 +252,7 @@ class Reactor (object):
 								peer = self.peers[key]
 								neighbor = peer.neighbor
 								# XXX: FIXME: Inet can only be compared to Inet
-								if connection.local == str(neighbor.peer_address) and (not neighbor.local_address or connection.peer == str(neighbor.local_address)):
+								if connection.local == str(neighbor.peer_address) and (neighbor.auto_discovery or connection.peer == str(neighbor.local_address)):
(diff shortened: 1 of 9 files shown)


lib/exabgp/reactor/loop.py
```diff
@@ -440,9 +440,8 @@ def restart (self):
 		self.processes.start()
 
 	def unschedule (self, peer):
-		key = peer.neighbor.name()
-		if key in self.peers:
-			del self.peers[key]
+		if peer in self.peers:
```
thomas-mangin: I am surprised by this change but if you are sure and tested it :-)

lib/exabgp/reactor/network/outgoing.py
```diff
@@ -21,22 +21,24 @@ def __init__ (self, afi, peer, local, port=179,md5='',ttl=None):
...
-			bind(self.io,local,afi)
-			async(self.io,peer)
-			connect(self.io,peer,port,afi,md5)
+				TTLv6(self.io, self.peer, self.ttl)
+			if local:
+				bind(self.io,self.local,afi)
+			async(self.io,self.peer)
+			connect(self.io,self.peer,port,afi,md5)
+			if not self.local:
+				self.local = self.io.getsockname()[0]
```
thomas-mangin: I think we should store this information in the neighbour configuration. in the local_address and have the fact it was auto-discovered in another variable (using for the printing of local-address auto/manual).

thomas-mangin:
It is a nice feature. Please, let me know what you think of my comment ?

lib/exabgp/reactor/loop.py
brijohn: This was done because with these changes neighbor.name() returns a different value for local_ip after we actually have connected and found our real ip address. Without this change we peers were not being unscheduled and this was causing problems.

lib/exabgp/reactor/network/outgoing.py
brijohn: Not quite sure what you mean here. Do you mean update the local_address field of neighbor with the auto discovered address and then have a separate field that we can query later for display purposes to tell us if it was auto-discovered or not?

brijohn:
I'll look into reworking this to work by setting local_address to auto for auto discovery.

I take it you are referring to MD5 auth peers that require you to bind to the same address they are configured to connect to?  

Auto discovery should work fine for that as long as there is only one IP that can be used to reach the peer in question, otherwise it could fail to establish a connection if the OS chooses the wrong IP.  I used Quagga as a base model for this, which allows you not set a specific update-source and it appears to work the same way from what I could tell.

thomas-mangin:
@brijohn thank you for accepting to change the patch. much appreciated.

I did not have MD5 in mind but the fact that a BGP session is setup by a tuple of IP which must be configured both side and must match. ExaBGP itself looks up the IP of the peer on the socket to find which peer is connecting and if there is a valid session setup for it.

I was trying to understand the use case but since yesterday I considered it could help with things like ansible templates where you do not want anything "local" specific in the configuration when announcing routes. If ExaBGP initiate the connection, the other end could accept a whole range (which is something ExaBGP does not do yet but I know other vendors support) and it would work.

thomas-mangin:
Hello @brijohn - the patch looks good, only one last thing if you do not mind: could we move the invalidation of MD5 within the configuration code. Quickly looking, I would guess there: https://github.com/Exa-Networks/exabgp/blob/master/lib/exabgp/configuration/neighbor/__init__.py#L151 and if it is done warn about it or perhaps better complain that the configuration is invalid.

thomas-mangin:
I will merge the patch and massage it, you can then review the outcome - seems simpler.

brijohn:
If you think that would be simpler I'm fine with it. I started work on some of the changes mentioned but haven't got through rerunning our tests on them yet,

thomas-mangin:
Sorry - I meant to do this work during the week-end and did not.

brijohn:
Thats ok. I added a few new commits to address your comments to this patch. 

It should now set local_address to None when configured for auto as well as set router id if its None and your connect time local address is an ipv4 address.

thomas-mangin:
merged - thanks

PR #599, opened 2017-04-11
Allow show routes commands to yield less often then every line

brijohn:
We were running into performance issues when sending exabgp a show routes command via the API.  The current code was yielding after printing each route. This works fine for smaller numbers of routes, but when you have multiple peers with a large number of routes stored in their RIB it was taking longer then desired to query the current set of routes.

This patch allows the setting of an environment variable that can configure how many lines will be printed before yielding back to the main loop, with the default being 1.

@thomas-mangin Feel free to suggest a better name for the environment variable; Naming is hard :-)

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/599)

### DIFF
--- a/lib/exabgp/configuration/setup.py
+++ b/lib/exabgp/configuration/setup.py
@@ -230,6 +230,12 @@ environment.configuration = {
 		},
 	},
 	'api':  {
+		'chunk': {
+			'read': environment.integer,
+			'write': environment.nop,
+			'value': '1',
+			'help': 'maximum lines to print before yielding in show routes api',
+		},
 		'encoder':  {
 			'read':  environment.api,
 			'write': environment.lower,
--- a/lib/exabgp/reactor/api/command/text.py
+++ b/lib/exabgp/reactor/api/command/text.py
@@ -18,7 +18,7 @@ from exabgp.bgp.message import OUT
 from exabgp.configuration.static import ParseStaticRoute
 
 from exabgp.version import version as _version
-
+from exabgp.configuration.environment import environment
 
 class Text (object):
 	callback = {}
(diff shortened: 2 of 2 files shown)


lib/exabgp/reactor/api/command/text.py
```diff
@@ -32,22 +32,26 @@ def register (function):
...
 		else:
 			peers = [n for n in reactor.peers.keys() if 'neighbor %s' % last in n]
+		line_count = 0
 		for key in peers:
 			peer = reactor.peers.get(key, None)
 			if not peer:
 				continue
 			if advertised:
 				families = peer._outgoing.proto.negotiated.families if peer._outgoing.proto else []
 			for change in list(peer.neighbor.rib.outgoing.sent_changes(families)):
```
thomas-mangin: The pythonic way to do this would be to have something like
```
routes = list(peer.neighbor.rib.outgoing.sent_changes(families))
while routes:
   changes,routes = changes[: lines_per_yield],changes[lines_per_yield:]
   for change in changes:
      ....
   yield True
```
However performance wise I do not know how it compares.

lib/exabgp/configuration/setup.py
```diff
@@ -230,6 +230,12 @@
 		},
 	},
 	'api':  {
+		'lines': {
```
thomas-mangin: not sure about the name neither .. 'api', 'grouping' but I can live with 'lines'

lib/exabgp/reactor/api/command/text.py
brijohn: Some quick testing doesn't show much if any performance difference between the two. So I'll switch this to use slices instead.  Both gave me in the range of 9-12 seconds to dump 60k routes (20k at a time). For comparison before this patch it takes 30-40 seconds.

lib/exabgp/configuration/setup.py
brijohn: Ack, So the name I meant to give it was chunk, as in this controls how the size of the chunk of data being sent back before yield is. Apparently instead I ended up calling the 'read' property 'chunk' instead.  This is what happens when i write code while tired :-)

So do you prefer, 'chunk', 'lines' or something else for the environment variable name?

thomas-mangin: patatoes, tomatoes .... I do not have a really better idea, happy to go with what you have.

PR #601, opened 2017-04-17
Drop duplicate routes during announce

brijohn:
When you announce an already announced route it gets inserted into the _modify_nlri and _modify_sorted dictionaries.  However it is also never cleared since that was only being done for actually announced routes, this can lead to increasing memory over time and in the case of re-announcing a large of routes decreased performance when sending announce route statements to the API. It was taking us more then 30 minutes to announce a set of 20k routes after hitting this bug.

This patch fixes this by moving duplicate route detection to the insert_announcement function and dropping them at that point.  It also fixes a crash that was occurring when not using the adj-rib-out, due to trying to delete items from a hash during iteration.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/601)

### DIFF
--- a/lib/exabgp/rib/store.py
+++ b/lib/exabgp/rib/store.py
@@ -182,6 +182,14 @@ class Store (object):
 				if self.cache and change_nlri_index not in self._seen.get(change.nlri.family(),{}):
 					return
 
+		if not force and self.cache:
+			announced = self._seen.get(change.nlri.family(),{})
+			if change_nlri_index in announced and change.nlri.action == OUT.ANNOUNCE:
+				old_change = announced[change_nlri_index]
+				# it is a duplicate route
+				if old_change.attributes.index() == change.attributes.index() and old_change.nlri.nexthop.index() == change.nlri.nexthop.index():
+					return
+
 		# add the route to the list to be announced
 		dict_sorted.setdefault(change_attr_index,{})[change_nlri_index] = change
 		dict_nlri[change_nlri_index] = change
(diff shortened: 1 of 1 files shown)


lib/exabgp/rib/store.py
```diff
@@ -182,6 +182,14 @@ def insert_announced (self, change, force=False):
 				if self.cache and change_nlri_index not in self._seen.get(change.nlri.family(),{}):
 					return
 
+		if not force and self.cache:
+			announced = self._seen.get(change.nlri.family(),{})
+			if change_nlri_index in announced and change.nlri.action == OUT.ANNOUNCE:
+				old_change = announced[change_nlri_index]
+				# it is a duplicate route
+				if old_change.attributes.index() == change.attributes.index() and old_change.nlri.nexthop.index() == change.nlri.nexthop.index():
```
thomas-mangin: The route in announced could be a withdrawal, if it is a withdrawal, then we need to just replace it with this new announce.
So I guess the test should be updated to:
```if old_change.nlri.action == OUT.ANNOUNCE  and old_change.attributes.index() == change.attributes.index() and old_change.nlri.nexthop.index() == change.nlri.nexthop.index():
```
what do you think ?

brijohn: Perhaps I missed something here, but isn't announced our adjacent outgoing rib? So shouldn't all the routes in it be routesd we are announcing? Any withdraws nlris that we have withdrawn would have been removed from announced when the route was withdrawn, correct?

thomas-mangin: It is my recollection (but I did not check the code) that multiple announce/withdraw command can be processed before this code is run. Therefore should someone flap a route fast enough, the code must be able to handle it.

brijohn: So from my reading of the code starting at line 247 in rib/store.py it looks like we only ever set a change entry in announced that has an action of OUT.ANNOUNCE, if we process an OUT.WITHDRAW change we actually remove that nlri from announced. Wouldn't this mean that checkout the action of the old_change would always be OUT.ANNOUNCE?

```
if self.cache:
    announced = self._seen
    for change in changed:
        if change.nlri.action == OUT.ANNOUNCE:
            announced.setdefault(change.nlri.family(),{})[change.index()] = change
        else:
            family = change.nlri.family()
            if family in announced:
                announced[family].pop(change.index(),None)
```

thomas-mangin: That looks more like the code you changed - I would still change the else to an `elif` testing for withdrawal as I believe the value can be `None` (it should never happen - but better explicit than implicit) ...

thomas-mangin: ok - seems right - committing.

PR #602, opened 2017-04-21
Add support base64 encoding the md5 password

brijohn:
This patch adds an option to support using base64 encoding for the md5 password. 

We use this to prevent issues with md5 passwords that might contain characters, like ;,  {, } or ". Even if we quote the md5 password in the config file there will still be issues if someone passes in a string containing quote marks.

By being able to set a base64 encoded version of the password we no longer have to worry about having exabgp crash if someone sets a password containing such characters.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/602)


thomas-mangin:
Thank you for this feature 👍

PR #616, opened 2017-05-09
SPEC File version bump and Requires format fix

lukehinds:
This change bumps the spec file to the latest release, and sets the `Requires` directives to seperate lines.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/616)


lukehinds:
Sorry, don't merge..I just noticed something else needs a fix, so will follow up with a single patch.

PR #619, opened 2017-05-09
batch of py3 changes

tmmorin:
This stack of changes is mostly py3 changes.

There is at the bottom, a commit to rever the ord/chr/concat_strs rename, and this reverted rename is reverted by a change at the end of the stack of commits. This was the easiest I found to efficiently rebase my stack of commits.

There are also a few cleanup changes, unrelated to py3.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/619)


thomas-mangin:
I think I manually ported all these change to master ... hopefully correctly ...

tmmorin:
Thanks for merging.

There still remains an issue: many call to concat_bytes() for which the argument is a list or a generator, while the function, as defined, requires an expansion.

I will push a commit to solve that.

tmmorin:
There were in fact quite a bunch of things that got lost in translation, hence the new PR with more, which also includes things I had missed in my initial submission.

Since I possibly revert things you had decided not to merge as is, please check...

thomas-mangin:
Can we not check the length in the function and if the length is zero handle the data differently - not looked at how it behaves so it may be a bad idea (and clearly not as fast as a dedicated correct function).

PR #624, opened 2017-05-10
fixes to the set of ported py3 patches

tmmorin:
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/624)

### DIFF
--- a/lib/exabgp/bgp/message/update/nlri/cidr.py
+++ b/lib/exabgp/bgp/message/update/nlri/cidr.py
@@ -72,7 +72,7 @@ class CIDR (object):
 		return "%s/%s" % (self.top(),self.mask)
 
 	def index (self):
-		return character(self.mask) + str(self._packed[:CIDR.size(self.mask)])
+		return chr(self.mask) + str(self._packed[:CIDR.size(self.mask)])
 
 	def pack_ip (self):
 		return self._packed[:CIDR.size(self.mask)]
--- a/lib/exabgp/util/__init__.py
+++ b/lib/exabgp/util/__init__.py
@@ -16,7 +16,7 @@ def hexstring (value):
 	def spaced (value):
 		for v in value:
 			yield '%02X' % ord(v)
-	return '0x' + concat_bytes(spaced(value))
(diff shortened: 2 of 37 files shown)


lib/exabgp/bgp/message/update/nlri/cidr.py
```diff
@@ -72,7 +72,7 @@ def prefix (self):
 		return "%s/%s" % (self.top(),self.mask)
 
 	def index (self):
-		return character(self.mask) + str(self._packed[:CIDR.size(self.mask)])
+		return chr(self.mask) + str(self._packed[:CIDR.size(self.mask)])
```
thomas-mangin: why chr() here please ?

lib/exabgp/util/__init__.py
```diff
@@ -58,10 +58,16 @@ def padding(n):
 def concat_strs(*items):
 	return ''.join(items)
 
-
 def concat_bytes(*items):
 	return b''.join(items)
 
+# same with iterators/lists
+
+def concat_strs_i(iterable):
```
thomas-mangin: I would rather not have to worry about the data type and have a generic function (as long as the performance are not widely affected) - what is your opinion ?

thomas-mangin:
Thank you for all that work ...

lib/exabgp/bgp/message/update/nlri/cidr.py
tmmorin: I think that index returns a string rather than a bytestring, and that we need a chr rather than a byte to allow concatenation with str(self._packed ...).

lib/exabgp/util/__init__.py
tmmorin: I would prefer merging as is, if only to unbreak master.

Then we can consider merging both under a same function, based on testing if there is only one argument which can be iterated over... is this what you have in mind ?

lib/exabgp/bgp/message/update/nlri/cidr.py
thomas-mangin: understood

PR #637, opened 2017-05-18
AIGP: Add missing concat import

ruissalo:
I came across this missing import. Looks like it's from this change
9eb878be (Thomas Morin  2017-05-10 00:13:41 -0400  55)

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/637)


thomas-mangin:
Thank you

PR #646, opened 2017-05-30
Global listening address changes

brijohn:
This adds support for configuring multiple bind address in the environment file for global listeners. Particularly we use this so we are able to setup for an IPv6 listener on :: and an IPv4 listener on 0.0.0.0. It also adds support for adding and removing the correct md5 key mappings to any configured global listeners when the configuration file changes so that incoming md5 connections will work.

This patch should only affect the global listeners defined in the environment file and not the per neighbor ones that can be created by the listen directive.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/646)

### DIFF
--- a/lib/exabgp/reactor/listener.py
+++ b/lib/exabgp/reactor/listener.py
@@ -52,17 +52,15 @@ class Listener (object):
 				continue
 			if local_port != port:
 				continue
-			if md5:
-				MD5(sock,peer_ip.top(),0,md5,md5_base64)
+			MD5(sock,peer_ip.top(),0,md5,md5_base64)
 			if ttl_in:
 				MIN_TTL(sock,peer_ip,ttl_in)
 			return
 
 		try:
 			sock = self._new_socket(local_ip)
-			if md5:
-				# MD5 must match the peer side of the TCP, not the local one
-				MD5(sock,peer_ip.top(),0,md5,md5_base64)
+			# MD5 must match the peer side of the TCP, not the local one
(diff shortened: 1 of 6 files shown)


lib/exabgp/reactor/listener.py
```diff
@@ -52,19 +52,21 @@ def listen (self, local_ip, peer_ip, local_port, md5, md5_base64, ttl_in):
 				continue
 			if local_port != port:
 				continue
-			if md5:
-				MD5(sock,peer_ip.top(),0,md5,md5_base64)
+			MD5(sock,peer_ip.top(),0,md5,md5_base64)
```
thomas-mangin: Hi @brijohn - why was the if test removed here ? accident ? it work without it ?

lib/exabgp/reactor/listener.py
```diff
@@ -52,19 +52,21 @@ def listen (self, local_ip, peer_ip, local_port, md5, md5_base64, ttl_in):
...
 			if ttl_in:
 				MIN_TTL(sock,peer_ip,ttl_in)
 			return
 
 		try:
 			sock = self._new_socket(local_ip)
-			if md5:
-				# MD5 must match the peer side of the TCP, not the local one
-				MD5(sock,peer_ip.top(),0,md5,md5_base64)
+			# MD5 must match the peer side of the TCP, not the local one
```
thomas-mangin: Same question for the test here too :-)

thomas-mangin:
And again, thank you for taking the time to write these patches - your help is much appreciated.

PR #672, opened 2017-07-04
Fix healthcheck.py to use /var/run/log for syslog on NetBSD

he32:
This allows it to work on NetBSD too.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/672)

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -203,10 +203,12 @@ def setup_logging (debug, silent, name, syslog_facility, syslog):
     """Setup logger"""
 
     def syslog_address():
-        """Return a sensitive syslog address"""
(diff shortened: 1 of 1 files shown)


he32:
Insisting on a list of OSes here probably isn't the best strategy, but I know
too little python to make a constructive suggestion for an improvement.

thomas-mangin:
Thank you

thomas-mangin:
It merged in a branch it seems - porting to 3.4 and master

vincentbernat:
Yes, that's unfortunate but Python doesn't abstract that. By default, it uses UDP on 514, but on many systems, the syslog daemon doesn't listen on the UDP port. A loop to test all possibilities, regardless of the OS, could also be used (check if file exists and is a socket).

PR #689, opened 2017-08-02
Update processes.py

songhu:
Throw IO error when line is empty instead of raw

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/689)

### DIFF
--- a/lib/exabgp/reactor/api/processes.py
+++ b/lib/exabgp/reactor/api/processes.py
@@ -196,7 +196,7 @@ class Processes (object):
 							line = raw.rstrip()
 							consumed_data = True
 							self.logger.processes("Command from process %s : %s " % (process,line))
-							if raw == '':
+							if line == '':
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Hello @songhu,

Obviously you provided this patch as something must be not as I expected but I would welcome some information on what `raw` looks like as AFAIK it should be the same as `line` when there is no data.

Also it would make sense to move the line before the logging entry if we are going to change this code ... Not sure why I missed that before !

thomas-mangin:
@songhu ping ? please ?

thomas-mangin:
@songhu if the goal is to allow empty line to be accepted via the API ( `\r`, `\n`, `\r\n`, `\s+\n` ,etc) then a new test should be added which cause ExaBGP to ignore input when `line` is `` but raw is not ... I would accept such change.

PR #692, opened 2017-08-07
fix #690: Add L2VPN/EVPN to next-hop sizing check

### LINKED ISSUE #690: Add L2VPN/EVPN to next-hop sizing check to fix L2VPN/EVPN peers
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Commit 4aed9cbf81cbe544253e7da7c97feeb238bfea43 added next-hop sizing checks but did not add the sizing information for L2VPN / EVPN address-family.  This causes L2VPN/EVPN peers to be torn down.

##### STEPS TO REPRODUCE

Connect to a L2VPN/EVPN neighbor and watch the session come down.

##### FIX

I corrected this with just a quick add to the Family class - but given the rest of the changes in the commit there may be other areas that need some attention for EVPN.

Without this - an EVPN peer gets torn down (with an unsupported message being sent) , after this adjustment the peer is stable and passing information.

```
diff --git a/lib/exabgp/protocol/family.py b/lib/exabgp/protocol/family.py
index 62b90a3..2747811 100644
--- a/lib/exabgp/protocol/family.py
+++ b/lib/exabgp/protocol/family.py
@@ -264,6 +264,7 @@ class Family (object):
                (AFI.ipv6,SAFI.flow_ip):   ((0,16,32), 0),
                (AFI.ipv6,SAFI.flow_vpn):  ((0,16,32), 0),
                (AFI.l2vpn,SAFI.vpls):     ((4,),      0),
+               (AFI.l2vpn,SAFI.evpn):     ((4,),      0),
                (AFI.bgpls,SAFI.bgp_ls):   ((4,),      0),
        }
```

(for small items like this I just copy n paste - but if a full merge req. is preferred - just let me know)

tmmorin:
Credits to jsynack who provided the solution in the bug report.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/692)

### DIFF
--- a/lib/exabgp/protocol/family.py
+++ b/lib/exabgp/protocol/family.py
@@ -264,6 +264,7 @@ class Family (object):
 		(AFI.ipv6,SAFI.flow_ip):   ((0,16,32), 0),
 		(AFI.ipv6,SAFI.flow_vpn):  ((0,16,32), 0),
 		(AFI.l2vpn,SAFI.vpls):     ((4,),      0),
+		(AFI.l2vpn,SAFI.evpn):     ((4,),      0),
 		(AFI.bgpls,SAFI.bgp_ls):   ((4,),      0),
 	}
 


tmmorin:
I believe that allowing the length tuple currently stored in Family.size to be provided in the NLRI.register hook would be nice, and would avoid future case where a new afi/safi is added, but actually fails because of size checks.

thomas-mangin:
Yes, it is a good idea. Happy to accept a patch doing it :-)

PR #705, opened 2017-08-16
Fix busy spin when receiving commands on named pipe

### LINKED ISSUE #704: exabgpcli makes exabgp pin 100% CPU
##### ISSUE TYPE

- Bug Report

##### OS
- Debian 9
- CentOS 7

##### VERSION
```
root@exabgp:~# INTERPRETER=python /mnt/hgfs/devtools/exabgp/sbin/exabgp --version
ExaBGP : 4.0.2-1c737d99
Python : 2.7.13 (default, Jan 19 2017, 14:48:08)  [GCC 6.3.0 20170118]
Uname  : Linux exabgp 4.9.0-3-amd64 #1 SMP Debian 4.9.30-2+deb9u3 (2017-08-06) x86_64
Root   : /mnt/hgfs/devtools/exabgp
```

##### ENVIRONMENT 

```
root@exabgp:~# INTERPRETER=python /mnt/hgfs/devtools/exabgp/sbin/exabgp --di

[exabgp.daemon]
drop = false
user = 'root'

[exabgp.tcp]
bind = '[0.0.0.0]'
```

##### CONFIGURATION
```
template {
  neighbor all {
    router-id 192.168.3.9;
    local-as 64400;
    hold-time 20;
    capability {
      graceful-restart;
    }
  }
}

process foo {
  run '/mnt/hgfs/devtools/exabgp/bin/showadjrib.py';
  encoder text;
}

neighbor 192.168.3.8 {
 inherit all;
  local-address 192.168.3.9;
  family {
    ipv4 unicast;
  }
  peer-as 64300;
}
```

##### SUMMARY

when using any command with `exabgpcli`, the exabgp process will start busy spinning the cpu.

##### STEPS TO REPRODUCE
Given the previous config and a peer configured properly, sending any command (valid or not) with `exabgpcli`:
```
exabgpcli help
```

##### EXPECTED RESULTS

`exabgp` should not pin a CPU

##### ACTUAL RESULTS

exabgp pins a CPU at 100%

##### IMPORTANCE
`exabgpcli` is pretty much useless as it would potentially prevent prevent exabgp from properly operating.

chantra:
When opening the read named pipe in RD_ONLY, the select loop will always
return that something is available to read from self.r_pipe once
something has opened and closed the fifo like `exabgpcli` or `echo "foo" >
/path/to/fifo`. Unfortunately, those are going to be EOF and exabgp will
be spinning trying to read content from the fifo when nothing is
actually available.
To work around that, we can open the reader fifo in RDWR mode, this will
ensure that the pipe will always have a potential writer and no EOF will
be available for select.

Fixes #704

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/705)

### DIFF
--- a/lib/exabgp/application/control.py
+++ b/lib/exabgp/application/control.py
@@ -134,7 +134,7 @@ class Control (object):
 
 	def loop (self):
 		try:
-			self.r_pipe = os.open(self.recv, os.O_RDONLY | os.O_NONBLOCK | os.O_EXCL)
+			self.r_pipe = os.open(self.recv, os.O_RDWR | os.O_NONBLOCK | os.O_EXCL)
 		except OSError:
 			self.terminate()
 
@@ -217,7 +217,7 @@ class Control (object):
 		}
 
 		def consume (source):
-			if not backlog and b'\n' not in store:
+			if not backlog[source] and b'\n' not in store[source]:
 				store[source] += read[source](1024)
 			else:
 				backlog[source].append(read[source](1024))


lib/exabgp/application/control.py
```diff
@@ -217,7 +217,7 @@ def fifo_writer (line):
 		}
 
 		def consume (source):
-			if not backlog and b'\n' not in store:
+			if not backlog[source] and b'\n' not in store[source]:
```
chantra: I fixed this here, as I assume this was wrong, unless you were explicitly looking for \n in store keys :)

chantra:
So... this patch fixes the busy spin on exabgp side. The `exabgpcli` is slow on linux, like order or 1.5 sec to run while it is instant on MacOSX.

While it will be nice to optimize, it is much less important than fixing the busy spin and remains functional.

chantra:
actually, the slow cli only reproduce for me on Debian 9 in a VM. This diff applied against 4.0.2 and ran on CentOS7 baremetal/container does not hang.

thomas-mangin:
Thank you very much for the fix another user was affected .. Yes, python is suppose to be 'code once, run everywhere' but like Java for some things ... well, you find differences between OSes.

My pure guess would be that the sleep time of the select function is handled differently perhaps as no real time clock is available on a VM ...

Issue #709, opened 2017-08-23
Route refresh slow on master


dhammika:
##### ISSUE TYPE

 - Bug Report

##### OS
Linux/Debian 8.9

##### VERSION

```
ExaBGP : 4.0.2-1c737d99
Python : 3.4.2 (default, Oct  8 2014, 10:45:20)  [GCC 4.9.1]
```

##### ENVIRONMENT 

```
```

##### CONFIGURATION
```
process morebacon {
	encoder json;
	run /usr/bin/morebacon;
}

neighbor 10.1.1.2 {
    description "10.1.1.2";
    router-id 10.1.1.1;
    local-address 10.1.1.1;
    peer-as 666;
    local-as 666;
    hold-time 720;
    group-updates;
    capability {
	    asn4 enable;
	    add-path disable;
	    route-refresh enable;
    }
    family {
	    ipv4 unicast;
    }
    api {
	    processes [morebacon];
	    neighbor-changes;
	    receive {
	        parsed;
	        update;
	    }
    }
}
```

##### SUMMARY

Route refresh takes considerably longer on master, as much as 10x compared to 3.4 branch.

##### STEPS TO REPRODUCE
Peer with a router as ibgp neighbor.

##### EXPECTED RESULTS

Exa reactor loop shouldn't block on polling forked processes with pending peer messages.

##### ACTUAL RESULTS

Reactor loop seems to be blocking 10ms (select sleep below) on each iteration with pending upstream messages. fd `#6` is forked process stdout pipe and fd `#4` is peer tcp connection.
```
write(5, "{ \"exabgp\": \"4.0.1\", \"time\": 150"..., 722) = 722
wait4(1956130, 0x7fff2ecdb6b4, WNOHANG, NULL) = 0
select(7, [6], [], [], {0, 0})          = 0 (Timeout)
select(7, [6], [], [], {0, 10000})      = 0 (Timeout)
select(5, [4], [], [], {0, 0})          = 1 (in [4], left {0, 0})
recvfrom(4, "\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\x00\x66\x02", 19, 0, NULL, NULL) = 19
select(5, [4], [], [], {0, 0})          = 1 (in [4], left {0, 0})
[...]

PR #710, opened 2017-08-23
Fix api poller #709

dhammika:

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/710)


thomas-mangin:
not merging as it but fixed with the same logic (just in a slightly more pythonic way).
Thank you very much for tracing this issue. Much appreciated.

PR #734, opened 2017-10-13
Remove Debian packaging from master branch

vincentbernat:
Provide instructions on how to build using the debian/sid branch. The
drawback of such an approach is that it may not work on older versions
of Debian/Ubuntu. If this happens, it would be possible to maintain a
derivative branch for such versions. This will still be easier to
maintain.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/734)


thomas-mangin:
Thank you Vincent.

PR #735, opened 2017-10-17
python-setuptools is a runtime requirement

feld:
Without python-setuptools the exabgp software built from python-exabgp.spec.git refuse to run with the error:

ImportError: No module named pkg_resources

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/735)


thomas-mangin:
Not building RPMS myself so I will take your word on it .. Thank you for noticing, reporting and providing a patch for the issue.

PR #740, opened 2017-10-24
Add a systemd exabgp instance service

jmauro:
Hello,

I was wondering if adding the systemd instance feature could be interesting when using exabgp? So I created the following PR to see if you, guys think this is a good idea or not.

The idea is pretty much the same as the `rsync` in Red Hat, the package delivers both the default rsyncd service and also the socket one, so I though this could be applied to exabgp.

Let me know if this does make sense or not.

Regards,
JM

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/740)

### DIFF
--- a/etc/systemd/exabgp@.service
+++ b/etc/systemd/exabgp@.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=ExaBGP for instance %i
+Documentation=man:exabgp(1)
+Documentation=man:exabgp.conf(5)
+Documentation=https://github.com/Exa-Networks/exabgp/wiki
+After=network.target
+ConditionPathExists=/etc/exabgp/exabgp-%i.conf
+
+[Service]
+Environment=exabgp_daemon_daemonize=false
+Environment=EXABGP_INSTANCE=exabgp-%i
+ExecStart=/usr/sbin/exabgp /etc/exabgp/${EXABGP_INSTANCE}.conf
+ExecReload=/bin/kill -USR1 $MAINPID
+
+[Install]
+WantedBy=multi-user.target
--- a/redhat/python-exabgp.spec
+++ b/redhat/python-exabgp.spec
@@ -50,6 +50,7 @@ mv ${RPM_BUILD_ROOT}/usr/etc/exabgp ${RPM_BUILD_ROOT}/%{_sysconfdir}/
 
 install -d %{buildroot}/%{_unitdir}
 install etc/systemd/exabgp.service %{buildroot}/%{_unitdir}/
+install etc/systemd/exabgp@.service %{buildroot}/%{_unitdir}/
 
 install -d %{buildroot}/%{_mandir}/man1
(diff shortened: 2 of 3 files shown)


vincentbernat:
I like the idea. Maybe use `/etc/exabgp/exabgp-%i.conf`?

Also, currently, the way ExaBGP searches for named pipes for CLI is not compatible with multiple instances. I suggest to pass the instance name in an environment variable and make ExaBGP uses `exabgp-%i.in` and `exabgp-%i.out` when this variable is present. But this can be done in a separate PR when someone needs it.

thomas-mangin:
@vincentbernat well spotted :-)
@jmauro if you do not mind fixing this issue I would happily merge

jmauro:
Sounds good will do today.

jmauro:
Hello @thomas-mangin,

Just push the new review. Let me know if this does what you want.

Regards,
JM

thomas-mangin:
@vincentbernat I am systemd user only .. if you could please do the code review I would appreciate. If you are happy, happy to merge.

vincentbernat:
It is OK for the systemd part but it is missing the "make ExaBGP uses `exabgp-%i.in` and `exabgp-%i.out` when this variable is present" part. We could merge now and wait for someone to request both functionalities at the same time.

thomas-mangin:
@jmauro would it be possible for you to add this variable to make sure that two exabgp do not use the same pipe .. that would really mess up both instance if it was going to happen ...

jmauro:
Hello @thomas-mangin would love to do it, but I am not sure I understand what you want me to do. Could you be a bit more specific or point me toward some documentation?

thomas-mangin:
@jmauro ExaBGP 4.0 can be controlled using exabgpcli with two named pipe. The name can be controlled via the environment variable. So to prevent name clash when multiple exabgp are running explicitly setting this value would make sense.
```
$ ./sbin/exabgp --help | grep pipe
 - exabgp.api.cli               : should we create a named pipe for the cli. default (true)
 - exabgp.api.pipename          : name to be used for the exabgp pipe. default ('exabgp')
```

jmauro:
Hello,

I have not forgotten, just didn't have time to work on it. Try to do it asap. 

Regards.,
JM

jmauro:
Hello @thomas-mangin,

Quick question for you, currently you can change the pipename inside `control.py` but this is not reflected into the `bgp.py` (cf: https://github.com/Exa-Networks/exabgp/blob/master/lib/exabgp/application/bgp.py#L66), so do you also want me to fix this?

Regards,
JM

thomas-mangin:
Oops yes, with pleasure if it is not the case -- but I thought I was reading the env variable in both program - will check.

thomas-mangin:
@jmauro should I merge as it or do you still want to do some changes ?

jmauro:
I am good with it, so yes you can merge 😄

PR #744, opened 2017-11-17
EVPN support for VXLAN

tmmorin:
In EVPN/VXLAN [1] a VXLAN VNI is stored inside the 3-byte fields used
for MPLS labels in RFC7432. In this case the 4-bit shifting done for
MPLS labels does not apply anymore. This applies to label fields in NLRI
routes and to the label field in the PMSI tunnel attribute.

This change updates corresponding classes to:
- expose both the MPLS 4-bit shifted value, and the raw value, in
  parsed routes
- allow passing a 'raw_label' value in routes to be encoded,
  when provided this value will be used for packing,
  instead of using the 'label' after 4-bit shifting
- update string representation to show both the 4-bit shifted label
  and the raw label

[1] https://tools.ietf.org/html/draft-ietf-bess-evpn-overlay

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/744)

### DIFF
--- a/lib/exabgp/bgp/message/update/attribute/pmsi.py
+++ b/lib/exabgp/bgp/message/update/attribute/pmsi.py
@@ -54,8 +54,9 @@ class PMSI (Attribute):
 
 	__slots__ = ['label','flags','tunnel']
 
-	def __init__ (self, tunnel, label, flags):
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Thank you a lot -- merged if you want to add a changelog entry for it too ?

Issue #743, opened 2017-11-11
BGP-LS parsing problem with IOS-XE


mancubus77:
##### ISSUE TYPE
 - Bug Report

##### OS
Red Hat Enterprise Linux Server release 6.9 (Santiago)

##### VERSION
```
| version       | 4.0.2-1c737d99
| interpreter   | 3.4.5 (default, Jun  1 2017, 13:52:39)  [GCC 4.4.7 20120313 (Red Hat 4.4.7-18)]
```

##### ENVIRONMENT 

```
14:07:57 | 9684   | welcome       | Thank you for using ExaBGP
14:07:57 | 9684   | version       | 4.0.2-1c737d99
14:07:57 | 9684   | interpreter   | 3.4.5 (default, Jun  1 2017, 13:52:39)  [GCC 4.4.7 20120313 (Red Hat 4.4.7-18)]
14:07:57 | 9684   | os            | Linux atlascol01 2.6.32-642.3.1.el6.x86_64 #1 SMP Sun Jun 26 18:16:44 EDT 2016 x86_64
14:07:57 | 9684   | installation  | /usr    
14:07:57 | 9684   | cli           | could not find the named pipes (exabgp.in and exabgp.out) required for the cli
14:07:57 | 9684   | cli           | we scanned the following folders (the number is your PID):
14:07:57 | 9684   | cli control   |  - /run/0/
14:07:57 | 9684   | cli control   |  - /run/
14:07:57 | 9684   | cli control   |  - /var/run/0/
14:07:57 | 9684   | cli control   |  - /var/run/
14:07:57 | 9684   | cli control   |  - /usr/run/0/
14:07:57 | 9684   | cli control   |  - /usr/run/
14:07:57 | 9684   | cli control   |  - /usr/var/run/0/
14:07:57 | 9684   | cli control   |  - /usr/var/run/
14:07:57 | 9684   | cli control   | please make them in one of the folder with the following commands:
14:07:57 | 9684   | cli control   | > mkfifo /var/home/stan/bgpls/run/exabgp.{in,out}
14:07:57 | 9684   | cli control   | > chmod 600 /var/home/stan/bgpls/run/exabgp.{in,out}
14:07:57 | 9684   | configuration | performing reload of exabgp 4.0.2-1c737d99
14:07:57 | 9684   | configuration | > process          | 'parsed-route-backend'
14:07:57 | 9684   | configuration | . run              | '/tmp/dump.py'
[...]

PR #748, opened 2017-11-30
Fix range() Py3 comp issue

ruissalo:
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/748)


ruissalo:
refixes #743

PR #762, opened 2017-12-28
Allow duplicate neighbor sections

### LINKED ISSUE #760: Duplicate peer definition may be non-fatal
##### ISSUE TYPE
 - Feature Idea

##### OS
Debian Linux

##### VERSION
```
ExaBGP : 4.0.2-1c737d99
Python : 3.6.4 (default, Dec 19 2017, 14:09:48)  [GCC 7.2.0]
Uname  : Linux zoro 4.14.0-1-amd64 #1 SMP Debian 4.14.2-1 (2017-11-30) x86_64
Root   :
```

##### CONFIGURATION
```
neighbor 80.81.195.174 {
 ...
}
neighbor 80.81.195.174 {
 ...
}
```

##### SUMMARY

Currently, when a neighbor is defined twice, we get a "duplicate peer definition" error. However, when ExaBGP is not using `listen`, there should be no problem in connecting to a peer twice. This could be done with a different `local-address` for each connection.

Here is a proposition to fix that:

 - accept a non-IP name for `neighbor`
 - `peer-address` defaults to name if this is an IP
 - ensure `peer-address` is present (either through name or directly)
 - modify unicity check to use name and not `peer-address`

Am I missing something?

vincentbernat:
They need to use a distinct name. In this case, `peer-address` needs
to be specified.

Currently not working! The `pre` part is not working as parsing doesn't trigger an exception. I have tried various other things, but I didn't quite find how to optionally parse the name as an IP address. If I remove `pre`, I don't get a name to parse. Any idea?

Would fix #760.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/762)

### DIFF
--- a/lib/exabgp/configuration/neighbor/__init__.py
+++ b/lib/exabgp/configuration/neighbor/__init__.py
@@ -127,7 +127,11 @@ class ParseNeighbor (Section):
 		self.neighbors = {}
 
 	def pre (self):
-		return self.parse(self.name,'peer-address')
+		try:
+			return self.parse(self.name,'peer-address')
+		except ValueError:
+			# That's OK
+			return True
 
 	def post (self):
 		for inherit in self.scope.pop('inherit',[]):
@@ -170,6 +174,9 @@ class ParseNeighbor (Section):
 		neighbor.multisession     = capability.get('multi-session',False)
 		neighbor.operational      = capability.get('operational',False)
 		neighbor.route_refresh    = capability.get('route-refresh',0)
+
+		if neighbor.peer_address is None:
+			return self.error.set('incomplete neighbor, missing peer address')
 
 		if capability.get('graceful-restart',False) is not False:
 			neighbor.graceful_restart = capability.get('graceful-restart',0) or int(neighbor.hold_time)
@@ -239,9 +246,9 @@ class ParseNeighbor (Section):
 		if neighbor.range_size > 1 and not neighbor.passive:
 			return self.error.set('can only use ip ranges for the peer address with passive neighbors')
 
-		if neighbor.peer_address.top() in self._neighbors:
-			return self.error.set('duplicate peer definition %s' % neighbor.peer_address.top())
-		self._neighbors.append(neighbor.peer_address.top())
+		if neighbor.name() in self._neighbors:
+			return self.error.set('duplicate peer definition %s' % neighbor.name())
+		self._neighbors.append(neighbor.name())
 
 		if neighbor.md5_password:
 			try:


lib/exabgp/configuration/neighbor/__init__.py
```diff
@@ -239,9 +246,9 @@ def post (self):
 		if neighbor.range_size > 1 and not neighbor.passive:
 			return self.error.set('can only use ip ranges for the peer address with passive neighbors')
 
-		if neighbor.peer_address.top() in self._neighbors:
-			return self.error.set('duplicate peer definition %s' % neighbor.peer_address.top())
-		self._neighbors.append(neighbor.peer_address.top())
+		if neighbor.name() in self._neighbors:
+			return self.error.set('duplicate peer definition %s' % neighbor.name())
+		self._neighbors.append(neighbor.name())
```
thomas-mangin: That's likely to break code - I would expect some code to rely on the IP address been here.

vincentbernat: I didn't find anything using the key of `self.neighbors` for anything else than iterating over it. And tests didn't trigger an error.

lib/exabgp/configuration/neighbor/__init__.py
```diff
@@ -171,6 +175,9 @@ def post (self):
 		neighbor.operational      = capability.get('operational',False)
 		neighbor.route_refresh    = capability.get('route-refresh',0)
 
+		if neighbor.peer_address is None:
+			return self.error.set('incomplete neighbor, missing peer address')
+
```
thomas-mangin: here we should check if we got the IP address from the name and if so assign it, it provide backward compatibility and allow the new behaviour ..

lib/exabgp/configuration/neighbor/__init__.py
```diff
@@ -239,9 +246,9 @@ def post (self):
 		if neighbor.range_size > 1 and not neighbor.passive:
 			return self.error.set('can only use ip ranges for the peer address with passive neighbors')
 
-		if neighbor.peer_address.top() in self._neighbors:
-			return self.error.set('duplicate peer definition %s' % neighbor.peer_address.top())
-		self._neighbors.append(neighbor.peer_address.top())
+		if neighbor.name() in self._neighbors:
+			return self.error.set('duplicate peer definition %s' % neighbor.name())
```
thomas-mangin: here we should go through all the peers and look at the src ip / dst ip / family defined, and in case of duplicate just warn the user - it should be enough to help them in case of mis-configuration but still allow your use case.

thomas-mangin:
To fix the issue you could check the tokeniser for data, with `self.tokeniser.line` and look if there is an IP or not, and if not generate a unique ID and add it to the scope .. something like `self.scope.to_context(unique_id_for_peer())`  

The reason I am not doing it myself is .. I do not remember how my own parsing code works most of the time, it takes me a few hours to re-understand it ...

I would _guess_ that the reason catching the exception does not work is that once the data is parsed it is added to the scope (a dict based class which represent the structure of the configuration) and the exception prevent this addition.

If you do not have time to experiment .. I will try in the incoming days.

vincentbernat:
Currently, I work around this limitation by running two ExaBGP. Since I only need this "feature" for a lab, this is fine for me. Therefore, I don't have much time to "fix" that. Don't spend time either if nobody else has never asked this feature. People can find the issue if they run into the same case.

thomas-mangin:
@vincentbernat Thank you for trying - when I can find some free time (instead of working week-ends) I will look into it.

Issue #660, opened 2017-06-20
Incorrect paths in python-exabgp.spec


wonderg:
https://github.com/Exa-Networks/exabgp/blob/3.4.19/redhat/python-exabgp.spec
Here this spec file has incorrect paths

```
# fix file locations
mv ${RPM_BUILD_ROOT}%{_bindir} ${RPM_BUILD_ROOT}%{_sbindir}
mv ${RPM_BUILD_ROOT}%{_sbindir}/healthcheck ${RPM_BUILD_ROOT}/%{_sbindir}/exabgp-healthcheck
install -d -m 744 ${RPM_BUILD_ROOT}/%{_sysconfdir}/
mv ${RPM_BUILD_ROOT}/usr/etc/exabgp ${RPM_BUILD_ROOT}/%{_sysconfdir}/
...
```

And CMD to build RPM at CentOS 7.2
`rpmbuild -ba /root/rpmbuild/SPECS/python-exabgp.spec`
raise an error 

```
Installing exabgp script to /root/rpmbuild/BUILDROOT/python-exabgp-3.4.18-1.el7.centos.x86_64/usr/bin
+ mv /root/rpmbuild/BUILDROOT/python-exabgp-3.4.18-1.el7.centos.x86_64/usr/bin /root/rpmbuild/BUILDROOT/python-exabgp-3.4.18-1.el7.centos.x86_64/usr/sbin
+ mv /root/rpmbuild/BUILDROOT/python-exabgp-3.4.18-1.el7.centos.x86_64/usr/sbin/healthcheck /root/rpmbuild/BUILDROOT/python-exabgp-3.4.18-1.el7.centos.x86_64//usr/sbin/exabgp-healthcheck
mv: cannot stat '/root/rpmbuild/BUILDROOT/python-exabgp-3.4.18-1.el7.centos.x86_64/usr/sbin/healthcheck': No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.Et1hOE (%install)
```

leleobhz:
Hello @wonderg 

i've created the .git version of this file and when I've created the .git spec, I've been hit by this healthcheck too. The git spec does compile well?

Thanks!

thomas-mangin:
happy to take a patch which fixes the issue but I did not write this and can not test it easily ...

wonderg:
@leleobhz yes, the same problem with python-exabgp.spec.git

@thomas-mangin if you could provide updated spec file here, then i could test it.
[...]

PR #770, opened 2018-01-15
Fix paths

wonderg:
This commit is a follow-up for https://github.com/Exa-Networks/exabgp/issues/660

Paths were taken from https://github.com/Exa-Networks/exabgp/blob/3.4.19/redhat/python-exabgp.spec.git

Current spec file was tested at CentOS72 and 74

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/770)


thomas-mangin:
Thank you - merged !

PR #774, opened 2018-01-21
Append related projects with `exa-template`

ton31337:
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/774)


thomas-mangin:
Thanks for sharing :-)

PR #783, opened 2018-02-11
Added support of range syntax in protocol configuration in Flows.

omrim9:
Added protocol validation (must be between 0-255)

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/783)

### DIFF
--- a/lib/exabgp/protocol/__init__.py
+++ b/lib/exabgp/protocol/__init__.py
@@ -26,6 +26,11 @@ class Protocol (Enum):
 	PIM   = 0x67
 	SCTP  = 0x84
 
+	def __init__(self, value):
+		super(Protocol, self).__init__()
+		if self < 0 or self > 255:
+			raise ValueError("Protocol must be between 0 and 255")
+
 	def pack (self):
 		return chr(self)
 
--- a/lib/exabgp/protocol/__init__.py
+++ b/lib/exabgp/protocol/__init__.py
@@ -27,7 +27,7 @@ class Protocol (Enum):
 	SCTP  = 0x84
 
 	def __init__(self, value):
-		super(Protocol, self).__init__()
+		Enum.__init__(self, value)
 		if self < 0 or self > 255:
 			raise ValueError("Protocol must be between 0 and 255")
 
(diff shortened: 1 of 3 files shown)


lib/exabgp/protocol/__init__.py
```diff
@@ -26,6 +26,11 @@ class Protocol (Enum):
 	PIM   = 0x67
 	SCTP  = 0x84
 
+	def __init__(self, value):
```
thomas-mangin: The constructor should not raise: 
 - if the data comes from the wire, the size will be acceptable and we use it (it could unknown)
 - if the data comes from the configuration file, the value should be sanitised at input not in the class

thomas-mangin:
Thank you for the change, the patch is good but the validation should be done at input time.
If you change it a little, I will be able to merge it "as it".

Could you also add an entry in `/CHANGELOG` for this feature with your name or GitHub account, please ?

omrim9:
Hi,
What do you mean at input time? Do you mean inside _tokenise() method of class Configuration?
Thanks

thomas-mangin:
I meant in the `def protocol` function, but as it is some "templated" code using `_generic_condition` which use the code below the solution is right .. I had not a chance to check it up.

The coding style does not use `super(Protocol, self).__init__()` but more the "old style"`Resource.__init__(self)` but the code is right.

```
class FlowIPProtocol (IOperationByte,NumericString,IPv4):
	ID  = 0x03
	NAME = 'protocol'
	converter = staticmethod(converter(Protocol.named,Protocol))
	decoder = staticmethod(decoder(ord,Protocol))
```

```
def converter (function, klass=None):
	def _integer (value):
		if klass is None:
			return function(value)
		try:
			return klass(value)
		except ValueError:
			return function(value)
	return _integer
```

omrim9:
Yes I found it to be the most suitable place to validate, because other places were a generic decorated code.
And if the data comes from the wire than it must be in a size of up to 1 byte, doesn't it? (at least that's what I found in IANA site of protocol numbers).
So do you want me to change the "super()" to "Enum.__init__(self, value)" and edit the changelog?
Thanks

thomas-mangin:
Yes, please and I will be able to Merge. I thought the parent class was Resource ..  but it is on 4.0 - could I also ask you to forward port the patch if 4.0 does not allow it already please ?

thomas-mangin:
Thank you !

PR #786, opened 2018-02-21
Support for sending a community when disabled in healthcheck.py

martbhell:
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/786)


vincentbernat:
:+1:

vincentbernat:
Hummm, after reading again, does ExaBGP accepts if you provide both `--community` and `--disabled-community`?

martbhell:
Good point. I'll try after lunch.

martbhell:
Yeah it broke stuff when one used both. Think I fixed it in #787 now.

PR #787, opened 2018-02-21
Make disabled-community work with community

martbhell:
Fix bug introduced in #786 which when one had both disabled-community and community in the healthcheck config then it added community [] twice to the announce line. Which crashed exabgp shortly after these errors:

<pre>
exabgp: out loop, skipping, not time yet
</pre>

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/787)

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -389,16 +389,26 @@ def loop (options):
                                                            options.next_hop or "self")
             if command == "announce":
                 announce = "{0} med {1}".format(announce, metric)
-                if options.community:
+                #With only disabled_community only add community string when we are down or disabled
+                if options.disabled_community and not options.community:
+                        if target in (states.DOWN, states.DISABLED):
+                            announce = "{0} community [ {1} ]".format(announce,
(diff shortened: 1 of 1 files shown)


martbhell:
Hmm.. I was a bit too early :/ 

Scenarios tested:

 - both disabled-community and community
   - normally: community appended
   - down and disabled: community changed to disabled-community
 - only community
   - normally, down and disabled: community appended
 - only disabled-community
  - no community on normal
  - no community added on down and disabled

vincentbernat:
Maybe, shorter code:

```python
if options.community or options.disabled_community:
    community = options.community
    if target in (states.DOWN, states.DISABLED) and options.disabled_community:
        community = options.disabled_community
    announce = "{0} community [ {1} ]".format(announce, community)
```

martbhell:
Nice! Added another commit that I tested also.

That shorter code almost worked, just needed another conditional to test that community was not None before adding it to the announce string.

vincentbernat:
Thanks! In master, there is support for extended and large communities, we'll see if someone wants an extended or large disabled community.

PR #798, opened 2018-03-23
healthcheck: fix next-hop handling

### LINKED ISSUE #797: healthcheck: next-hop argument should not be sent through ip_address()
healthcheck.py's `--next-hop` argument is specified with `type=ip_address`. However, ip_address() really constructs an IP network, so whatever's passed in will come out as a /32 (in the IPv4 case), which results in this API message (note the /32 on the next-hop):

```
DEBUG[healthcheck] exabgp: announce route 10.20.9.99/32 next-hop 10.20.13.48/32 med 100
announce route 10.20.9.99/32 next-hop 10.20.13.48/32 med 100
```

If I'm following the flow right, the main process then rejects this announce:

```
02:23:18 | 32     | reactor       | async | testing | problem with function
02:23:18 | 32     | reactor       | async | testing | illegal IP address string passed to inet_pton
```

Using the same constructor for both the `--ip` and `--next-hop` arguments seems like the wrong thing to do here.

vincentbernat:
next-hop must be an IP address, not a network. In a 8e432986f2d7, we
start accepting networks for advertised addresses and broke next-hop
handling in the process.

Fix #797

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/798)

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -58,19 +58,25 @@ logger = logging.getLogger("healthcheck")
 
 try:
     # Python 3.3+ or backport
-    from ipaddress import ip_network as _ip_address  # pylint: disable=F0401
+    from ipaddress import ip_network as ip_network  # pylint: disable=F0401
+    from ipaddress import ip_address as ip_address  # pylint: disable=F0401
 
-    def ip_address(x):
-        try:
-            x = x.decode('ascii')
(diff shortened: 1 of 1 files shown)


DrewBloechl:
Lightly tested, and seems to fix my issue. Thanks!

vincentbernat:
Bug is not present in 3.4, so no backport needed.

Issue #800, opened 2018-04-10
Withdrwals include attributes


dhammika:
##### Summary
Route withdrawals include default path attributes.
Withdrawals without mandatory nexthop don't strictly confirm with bgp spec and remote peer terminates the session with `Missing Well-known Attribute` error msg.

##### OS
```
Linux/Deb
```

##### Version
```
ExaBGP : master-e41160bce0da9bef05cb2abd56a74e9fc8e39a7b
Python : 3.5.2 (default, Nov 23 2017, 16:37:01)  [GCC 5.4.0 20160609]
```

##### Installation
```
git
```

##### Environment 

```

```

##### Configuration
```
neighbor 127.0.0.1 {
    router-id 127.0.0.2;
    local-address 127.0.0.2;
    local-as 12345;
    peer-as 12345;
    hold-time 30;
    group-updates;
    capability {
            asn4 enable;
            route-refresh enable;
    }
    family {
        ipv4 unicast;
    }
}
```

##### Program output
```
Withdrawal update dissect,

Internet Protocol Version 4, Src: 127.0.0.2, Dst: 127.0.0.1
Transmission Control Protocol, Src Port: 36792, Dst Port: 179, Seq: 100, Ack: 58, Len: 42
Border Gateway Protocol - UPDATE Message
    Marker: ffffffffffffffffffffffffffffffff
    Length: 42
    Type: UPDATE Message (2)
    Withdrawn Routes Length: 5
    Withdrawn Routes
        10.10.10.10/32
    Total Path Attribute Length: 14
    Path attributes
        Path Attribute - ORIGIN: IGP
        Path Attribute - AS_PATH: empty
        Path Attribute - LOCAL_PREF: 100

```

##### Steps to reproduce

```
Setup bgp peer and send a route withdrawal.
```

##### Importance
```
***Critical*** 
Immediate session flap on issuing the withdrawal msg.
```

dhammika:
@thomas-mangin any thoughts on this?

thomas-mangin:
Hi @dhammika - thanks for spotting this .. and sorry for the time it takes me to reply ATM. Patches applied, if required I could release a new version.

dhammika:
New version tag would be great, thanks!!

PR #801, opened 2018-04-10
Fix strip default attributes from withdrawals #800

dhammika:

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/801)


thomas-mangin:
@dhammika thanks for the patch - all good - missing an entry in the CHANGELOG - will add it.

PR #802, opened 2018-04-19
Add back in 3.x logic which stores existing routes over reload to fix…

longmalx:
I discovered that on 4.x, if I deleted a static route then did a SIGUSR1 or used the exabgpcli reload command, then the static route was not actually removed from the RIB or neighbor.

I tracked this down to the configuration reload refactoring.

A quick solution is to reinstate the old code which copied the changes over reload from 3.x - this doesn't seem the most elegant solution and I'm not clear whether the _previous_neighbors field was intended for something like this (albeit its logic is different).

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/802)

### DIFF
--- a/lib/exabgp/configuration/configuration.py
+++ b/lib/exabgp/configuration/configuration.py
@@ -62,6 +62,7 @@ class _Configuration (object):
 	def __init__ (self):
 		self.processes = {}
 		self.neighbors = {}
+		self.backup_changes = {}
 		self.logger = Logger ()
 
 	def inject_change (self, peers, change):
@@ -340,6 +341,8 @@ class Configuration (_Configuration):
 
 	def _clear (self):
 		self.processes = {}
+		for neighbor in self.neighbors:
+			self.backup_changes[neighbor] = self.neighbors[neighbor].changes
 		self.neighbors = {}
 		self._neighbors = {}
 		self._previous_neighbors = {}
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Sorry I started as review and just realised that I did not sent it .. This code seems to be a duplication of 
what should be done with `_previous_neighbors` so perhaps what is just needed is to save the state in `_previous_neighbors` as the restore part of the code is already there in that case.

What do you think ?

longmalx:
I agree the logic looks very close to `_previous_neighbors` and in fact I initially tried that, but assigning the `changes' in `_commit_reload' did not actually result in the correct behavior IIRC versus using the `backup_changes` field (I was keen to avoid changing the logic in `peer.py`).

If the intention though was the `_previous_neighbors` was for this use-case, then I agree that's the correct approach.

thomas-mangin:
@longmalx I do not comment the code enough 👎 but looking at it, it is indeed to carry neighbour information across reconfiguration so I have to assume it is indeed what it was for ... what issue did you experience when trying to use this variable vs `backup_change` ?

longmalx:
Let me try again and report back when I get a little spare time later today
or tomorrow.

longmalx:
I have switched the patch to use _previous_neighbors - initially testing shows this appears to work. I've added/removed static routes and verified that a reload is handled correctly, including when API routes exist as well.

thomas-mangin:
Super - will merge today and add a changelog entry

longmalx:
Thanks Thomas.
I've added a PR which fixes the RPM spec files - may be worth merging them before creating 4.0.6? See #803

PR #804, opened 2018-04-25
RPM spec files need to use 'next' version number rather than 'current'

longmalx:
This is a follow on from yesterday's PR which followed the previous Git RPM example of using `python setup.py current` to find the current version.

However, testing this morning on tags/4.0.6 shows that this actually gives the wrong version number! 
With the way `setup.py` currently works, `next` is the correct version to use.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/804)


longmalx:
Actually this is not correct yet - works for tags got `git rev-parse $(python setup.py next)` only works correctly for tagged branches.

longmalx:
Fixed now - seems sensible.

PR #806, opened 2018-04-28
qa: fix command to execute to replicate parsing error

vincentbernat:
Environment variables should use "_" instead of ".". We keep "-p" and
"-d" to get more results.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/806)


vincentbernat:
Is that expected Travis is not running anymore?

thomas-mangin:
both should work but _ is safer indeed - thanks.

PR #809, opened 2018-04-29
Fix some examples

vincentbernat:
Hey!

One example was missing an `inherit`. I've also added a second commit because it seems relative paths are now relative to the configuration file directory, not the current directory, right?

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/809)


thomas-mangin:
Thank you.

Issue #812, opened 2018-05-10
Reloading an invalid configuration breaks the API and CLI


longmalx:
##### Summary
If you attempt to reload a configuration which is invalid, then the processes are not reverted correctly.
This means that subsequent CLI commands will fail with the result "could not send command to ExaBGP".

Doing a reload via SIGUSR1 will restore functionality.

##### OS
bash-4.2# cat /etc/redhat-release 
CentOS Linux release 7.4.1708 (Core) 
bash-4.2# uname -a
Linux linuxvm 3.10.0-514.2.2.el7.x86_64 #1 SMP Tue Dec 6 23:06:41 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

##### Version
bash-4.2# exabgpcli version
exabgp 4.0.6-daa3d6ba

##### Installation
RPM build using the bundled SPEC files.

##### Environment 
bash-4.2# exabgp --di
bash-4.2# 

##### Configuration
n/a

##### Program output
Output when error occurs:
05:52:10 | 1448   | process         | command from process api-internal-cli-c10bcfd4 : reload 
05:52:10 | 1448   | process         | responding to api-internal-cli-c10bcfd4 : reload in progress
05:52:10 | 1448   | process         | responding to api-internal-cli-c10bcfd4 : done
05:52:10 | 1448   | configuration   | performing reload of exabgp 4.0.6-daa3d6ba
05:52:10 | 1448   | configuration   | > template         | 
05:52:10 | 1448   | configuration   | > neighbor         | 'rtbh-provider'
05:52:10 | 1448   | configuration   | . adj-rib-in       | 'fadfjhlse'
05:52:10 | 1448   | configuration   | problem with the configuration file, no change done
05:52:10 | 1448   | configuration   | 
05:52:10 | 1448   | configuration   | syntax error in section template/neighbor/rtbh-provider
05:52:10 | 1448   | configuration   | line 3: adj-rib-in fadfjhlse ;
05:52:10 | 1448   | configuration   | 
05:52:10 | 1448   | configuration   | invalid value (fadfjhlse) for a boolean
05:52:10 | 1448   | process         | terminating process api-internal-cli-c10bcfd4

##### Steps to reproduce
Start ExaBgp with valid configuration
Change configuration to be invalid
exabgpcli reload
exabgpcli version

```

thomas-mangin:
resolved by #813

PR #813, opened 2018-05-10
Fix handling of processes when the configuration reload fails

longmalx:
This appears to resolve the issue reported in #812 

I've made the handling of processes consistent with the reload success case.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/813)


thomas-mangin:
Added the changelog - thank you for your help tracking this bug ! It is much appreciated.

Issue #758, opened 2017-12-22
NOTIFY does not include peer information or raw body/header


spakka:
Hi Thomas,

##### ISSUE TYPE
 - Bug Report

##### OS
CentOS release 6.9 (Final)

##### VERSION

```
ExaBGP : 3.4.23
Python : 2.7.3 (f66246c46ca30b26a5c73e4cc95dd6235c966b8f, Jul 30 2013, 09:27:06) [PyPy 2.0.2 with GCC 4.4.7 20120313 (Red Hat 4.4.7-3)]
Uname  : #1 SMP Wed Jul 12 14:17:22 UTC 2017
```

##### ENVIRONMENT 

```
[exabgp.api]
highres = true

[exabgp.cache]
attributes = false
nexthops = false

[exabgp.log]
configuration = false
destination = 'syslog'
```

##### CONFIGURATION

```
    process receive-routes {
        encoder json;
	run /opt/exabgp/bin/queuer.py;
        neighbor-changes;
        receive {
            update;
            parsed;
            packets;
            consolidate;
            open;
            keepalive;
            notification;
        }
    }
```

##### SUMMARY
When Exabgp 3.4.23 receives a notify message from its peer, the received message does not include the 'neighbor' structure that contains the details about which peer the notify was sent from. It is also missing the 'body' and 'header' sections that we use to get the original raw message. These are present on Open, Update and Keepalive, but not Notify.

##### STEPS TO REPRODUCE
Configure a peering session towards a BGP neighbor that is not configured to accept the session.
Some implementations will accept the TCP connection, exchange OPEN messages, and then reply with a NOTIFY (error[Cease / Connection Rejected])

##### ACTUAL RESULTS
```
{
  "notification": {
    "subcode": 5,
    "code": 6,
    "data": "0605"
  },
  "counter": 531,
  "pid": "27552",
  "exabgp": "3.4.8",
  "host": "rrc20.ripe.net",
  "time": 1513869151.32,
  "ppid": "1",
  "type": "notification"
}
{
  "counter": 532,
  "pid": "27552",
  "exabgp": "3.4.8",
  "host": "rrc20.ripe.net",
  "neighbor": {
    "reason": "out loop, peer reset, message [notification received (6,5)] error[Cease / Connection Rejected]",
    "ip": "2001:7f8:24::47",
    "state": "down",
    "asn": {
      "peer": "34288",
      "local": "12654"
    },
    "address": {
      "peer": "2001:7f8:24::47",
[...]

PR #815, opened 2018-05-16
Include 'neighbor' structure when Exabgp receives a notify message.

andrekampert:
Fixes https://github.com/Exa-Networks/exabgp/issues/758


thomas-mangin:
ack - LGTM - but need an entry in the CHANGELOG.

PR #832, opened 2018-06-27
Reenable prefix for IPv4 ip-reachability-tlv

Tinusf:
I reenabled the prefix for ip-reachability-tlv.

### DIFF
--- a/lib/exabgp/bgp/message/update/nlri/bgpls/tlvs/ipreach.py
+++ b/lib/exabgp/bgp/message/update/nlri/bgpls/tlvs/ipreach.py
@@ -34,9 +34,10 @@ from exabgp.util import ordinal
 
 class IpReach(object):
 
-	def __init__ (self, prefix, packed=None):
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Thank you for this patch. Could you please change the patch so the JSON data is not modified please?  I would rather not break any existing parsers ? adding the length as an extra attribute would be fine.

lib/exabgp/bgp/message/update/nlri/bgpls/tlvs/ipreach.py
```diff
@@ -34,9 +34,10 @@
 
 class IpReach(object):
 
-	def __init__ (self, prefix, packed=None):
+	def __init__ (self, prefix, packed=None, plength=None):
```
thomas-mangin: The functions always take `packed` as the latest argument, could you please swap the order ?

Tinusf:
Ok I updated it to return two different attributes now.

PR #835, opened 2018-06-29
fix parsing of labelled default IP route

tmmorin:
The current check <24 is off by one and should be <=24,
which breaks parsing in the case of a prefix of lenth zero
(default route).

This change fixes this test.

This change also makes rd_mask a constant because the size
of this field does not depend on whether we are VPNv4 or VPNv6;
RD is always 8 bytes.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/835)

### DIFF
--- a/lib/exabgp/bgp/message/update/nlri/inet.py
+++ b/lib/exabgp/bgp/message/update/nlri/inet.py
@@ -26,6 +26,8 @@ from exabgp.bgp.message.update.nlri.qualifier import PathInfo
 from exabgp.bgp.message.update.nlri.qualifier import RouteDistinguisher
 from exabgp.bgp.message.notification import Notify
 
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Thank you Thomas.

thomas-mangin:
@tmmorin fixed the first but the code also break parsing of etc/exabgp/conf-prefix-sid.conf
So it is really really buggy. Can you please fix and then run ./qa/bin/parsing and ./qa/bin/functional all to make sure all is ok please or I will have to rollback

thomas-mangin:
In the meanwhile I have reverted on rd_mask is calculated - if not right please submit another PR.

tmmorin:
Your fix for rd_mask is correct, I had wrongly understood what this part of the code was doing.

PR #845, opened 2018-07-23
Send state down messages on peer removal

andrekampert:
When we remove peers from the configuration, we don't receive down messages for these peers, causing the announcements from those peers to still be visible. This fix explicitly sends down messages on removal.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/845)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Version explained:
  bug   : increase on bug or incremental changes
 
 Version 3.4.27
+ * Fix: Send state down messages on peer removal and shutdown
  * Improvement: RFC 7674 - Support redirect traffic to a VRF encoded as 4-octet AS, 2-octet Value
    patch by: Omri Matitiau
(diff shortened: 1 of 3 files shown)


thomas-mangin:
Thank you @andrekampert - LGTM - I am sure the patch got enough testing. Do you want to add an entry in the changelog so I can merge "as it" please ?

I will also need to forward port ...

andrekampert:
Hi @thomas-mangin - We tested the patch, including the new additions. Now we receive down messages in case of peer removal or shutdown. I've also updated the CHANGELOG. I can create a pull request for the forward port as well if you like, but I have no reliable means of testing that one.

thomas-mangin:
will merge once I have the same code on master. Hopefully this week.

thomas-mangin:
Ok, merged and forward ported, thank you very much for your help.

PR #846, opened 2018-07-27
Add header and body for notifications

andrekampert:
The header and body were missing for notifications in the 3.4 branch. They were already added to the master branch.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/846)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@ Version explained:
  bug   : increase on bug or incremental changes
 
 Version 3.4.27
+ * Fix: Add header and body for notifications
  * Fix: Send state down messages on peer removal and shutdown
  * Improvement: RFC 7674 - Support redirect traffic to a VRF encoded as 4-octet AS, 2-octet Value
    patch by: Omri Matitiau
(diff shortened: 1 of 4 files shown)


andrekampert:
@thomas-mangin Would you consider merging this pull request and create a new 3.4 release for us?

thomas-mangin:
Hello @andrekampert,

Sorry I missed your initial PR .. 

I am conflicted about merging it. It will change the format of the message sent and could potentially break some code parsing not expecting the extra data ..

I can see why you would like to see the behaviour changed and I am not sure that it is big enough a change to warrant a configuration option to toggle the data on or off..

Perhaps a middle ground would be to add the option, turn it on, but allow people to disable it if it break things and make it very clear in the release note that it may break some API code ...

andrekampert:
Dear @thomas-mangin 

I can see how this would be a breaking change, especially if some code makes assumptions about which messages have a header or body. Adding the option might be tricky as it seems like a reasonable default to use the new behaviour. At least, it would make it possible to revert to the old message format.

Note that this branch has another problem with the notification message, as the data section contains the full body, instead of the detail message. I decided not to change that to prevent breaking any code that relies on that full body being there.

Please let me know if there is anything I can do to improve this PR.

thomas-mangin:
@andrekampert sorry for leaving the patch in limbo for so long. I will stand by my initial assessment. Thank you for your help and sorry for not merging what was some good work!

Issue #848, opened 2018-09-11
Flow route parser doesn't support != operator


dhammika:
##### Summary

Flow route parser doesn't support != operator.

##### OS

```
Linux/Deb 9
```

##### Version

```
ExaBGP : master-904cd0c172a94cb1fff0f7c96f503f036ecce824
Python : 3.5.2 (default, Nov 23 2017, 16:37:01)  [GCC 5.4.0 20160609]
Uname  : Linux localhost 4.4.132-..
Root   : /home/d/dev/exabgp
```

##### Installation

```
git clone
```

##### Environment 

```
```

##### Configuration

```
```

##### Program output

```
16:01:02 | 20356  | configuration | performing reload of exabgp master-904cd0c172a94cb1fff0f7c96f503f036ecce824
16:01:02 | 20356  | reactor       | loaded new configuration successfully
16:01:53 | 20356  | api           | command could not parse flow in : announce flow route { match { source 10.0.0.1/32; destination 10.0.0.2/32; port [ !=443 ]; } then { discard; } }
16:01:53 | 20356  | api           | reason: unknown port !=443
```

##### Steps to reproduce

```
Setup ipv4 flow peer
echo 'announce flow route { match { source 10.0.0.1/32; destination 10.0.0.2/32; port [!=443]; } then { discard; } } > /tmp/exa.cmd
```

##### Importance

```
Lack of basic != operator severly limits flowspec rfc compliance.
```
https://tools.ietf.org/html/draft-ietf-idr-rfc5575bis-06#section-4.2.3

thomas-mangin:
was resolved with the merge above - closing.

PR #849, opened 2018-09-11
Fix flow route parser to handle != operator  #848

dhammika:

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/849)


thomas-mangin:
Hello @dhammika - seems sounds overall but changing the code does cause some of the test to fails so I will look into why.

dhammika:
Is it api-flow test?

Ordering of back to back withdraw and announce flow commands aren't strictly guaranteed? 
https://github.com/Exa-Networks/exabgp/blob/master/etc/exabgp/run/api-flow.run#L13

Also withdraw on L13 is only sending a empty-update without any withdraw NLRIs. 
`FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF:0025:02:0000000E4001010040020040050400000064`
https://github.com/Exa-Networks/exabgp/blob/master/qa/ci/api-flow.msg#L2

Is that correct though? We've seen some routers go bonkers on empty updates.

dhammika:
@thomas-mangin any thoughts on this? 
What test is failing, I can double check it.

thomas-mangin:
You can find out with `./qa/bin/functonal all` and then
`./qa/bin/functional server 22` in one terminal and 
`./qa/bin/functional client 22` in another one
(if test 22 fails for example)

thomas-mangin:
I would suggest the following changes to the patch
```
diff --git a/lib/exabgp/bgp/message/update/nlri/flow.py b/lib/exabgp/bgp/message/update/nlri/flow.py
index 825e5ed5..fb52048e 100644
--- a/lib/exabgp/bgp/message/update/nlri/flow.py
+++ b/lib/exabgp/bgp/message/update/nlri/flow.py
@@ -74,6 +74,7 @@ class CommonOperator (object):
 class NumericOperator (CommonOperator):
 	# reserved= 0x08  # 0b00001000
 	TRUE      = 0x00  # 0000000000
+	NEQ       = 0x06  # 0b00000110
 	LT        = 0x04  # 0b00000100
 	GT        = 0x02  # 0b00000010
 	EQ        = 0x01  # 0b00000001
@@ -243,7 +244,7 @@ class NumericString (object):
 		NumericOperator.EQ: '=',
 		NumericOperator.LT | NumericOperator.EQ: '<=',
 		NumericOperator.GT | NumericOperator.EQ: '>=',
-		NumericOperator.LT | NumericOperator.GT: '!=',
+		NumericOperator.NEQ: '!=',
 		NumericOperator.FALSE: 'false',

 		NumericOperator.AND: '&true',
@@ -252,7 +253,7 @@ class NumericString (object):
 		NumericOperator.AND | NumericOperator.EQ: '&=',
 		NumericOperator.AND | NumericOperator.LT | NumericOperator.EQ: '&<=',
 		NumericOperator.AND | NumericOperator.GT | NumericOperator.EQ: '&>=',
-		NumericOperator.AND | NumericOperator.LT | NumericOperator.GT: '&!=',
+		NumericOperator.AND | NumericOperator.NEQ: '&!=',
 		NumericOperator.AND | NumericOperator.FALSE: '&false',
 	}

diff --git a/lib/exabgp/configuration/flow/parser.py b/lib/exabgp/configuration/flow/parser.py
index 034ac59a..c5d2b524 100644
--- a/lib/exabgp/configuration/flow/parser.py
+++ b/lib/exabgp/configuration/flow/parser.py
@@ -97,12 +97,21 @@ def _operator_numeric (string):
 			operator = NumericOperator.GT
 		elif char == '<':
 			operator = NumericOperator.LT
-		elif char == 't' and string.lower().startswith('true'):
-			return NumericOperator.TRUE,string[4:]
-		elif char == 'f' and string.lower().startswith('false'):
-			return NumericOperator.FALSE,string[5:]
+		elif char == '!':
+			if string.startswith('!='):
+				return NumericOperator.NEQ,string[2:]
+			raise ValueError('invalid operator syntax %s' % string)
+		elif char == 't':
+			if string.lower().startswith('true'):
+				return NumericOperator.TRUE,string[4:]
+			raise ValueError('invalid operator syntax %s' % string)
+		elif char == 'f':
+			if string.lower().startswith('false'):
+				return NumericOperator.FALSE,string[5:]
+			raise ValueError('invalid operator syntax %s' % string)
 		else:
 			return NumericOperator.EQ,string
+
 		if string[1] == '=':
 			operator += NumericOperator.EQ
 			return operator,string[2:]
```

dhammika:
Incorporated your suggestions, but not sure about
```
+		elif char == 't':
+			if string.lower().startswith('true'):
+				return NumericOperator.TRUE,string[4:]
+			raise ValueError('invalid operator syntax %s' % string)
+		elif char == 'f':
+			if string.lower().startswith('false'):
+				return NumericOperator.FALSE,string[5:]
+			raise ValueError('invalid operator syntax %s' % string)
```
This will break `protocol tcp` or `protocol fc` type strings.

Did run functional tests and saw occasional failures in tests 09, 38 from update msg ordering.
Believe it's a timing issue on client flushing pending updates and was able to repro those on current master branch too.

thomas-mangin:
I possibly did a mistake adding the two extra raise ValueError .. Can you please remove them and I will decide what to do when I time please - sorry for that ! ...

dhammika:
Patch doesn't include those lines, pls take a look.

thomas-mangin:
Thank you .. By adding a test, we/I now need to add a test for this new rule:
https://github.com/Exa-Networks/exabgp/blob/master/qa/ci/conf-flow.msg
will do asap.

dhammika:
Added hex payload for update, pls take a look.

PR #852, opened 2018-10-11
Read pipename from env file rather than using default

craigmilne:
We'd like to be able to run two instances of ExaBGP on a single machine, supplying a different exabgp.env to each and have both use different pipenames. 

I noticed that this (or very similar) functionality was removed a year or so ago but it's not 100% clear to me why. Made sure this work when the .env file specifies the pipename value and leaves it absent.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/852)


thomas-mangin:
Thank you very much- LGTM.

PR #862, opened 2018-11-23
Update parse-multiple-process.conf

modir:
At the moment this file is equal to parse-process.conf. I guess the second process should have been added like my example now. Or if you intended to show something else with this file then please correct it yourself.

---
This change is [Reviewable](https://reviewable.io/reviews/exa-networks/exabgp/862)


thomas-mangin:
Thanks

PR #874, opened 2019-02-06
Adding option --local-preference to healtchcheck.py

hbo:
I propose to add option --local-preference to healthcheck.py.

The patch does this:
-  if not specified, the behavior of healthcheck.py is unchanged
- if healthcheck is called with --local-preference P, then the announcement towards the exabgp engine is extended with 'local-preference P'

Quite useful for single-IP announcements.

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -480,11 +480,9 @@ def loop(options):
                 options.next_hop or "self")
 
             if command == "announce":
-                localprefphrase=""
+                announce = "{0} med {1}".format(announce, metric)
                 if options.local_preference >= 0:
-                    localprefphrase = " local-preference {}".format(options.local_preference)
-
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Hello @hbo,

Thank you for this patch and for wanting to contribute to ExaBGP. Unfortunately I can not merge your patch yet as it contains an bug.

If you look at the line AFTER your change, you will see that the MED does rewrite the line
```
announce = "{0} med {1}{2}".format(announce, metric, localprefphrase)
```

Your code should use the same pattern, could you please update it ?

Thank you.

hbo:
Hi @thomas-mangin ,

I simplified the patch. Much better this way, you are right.

```
+                if options.local_preference >= 0:
+                    announce = "{0} local-preference {1}".format(announce, options.local_preference)
```
Regards

thomas-mangin:
@vincentbernat LGTM - fine to merge ?

vincentbernat:
Yes, looks good to me too!

Issue #875, opened 2019-02-17
error on ranged neighbor peer connection: AttributeError: 'Listener' object has no attribute 'processes'


wenxin-wang:
##### Summary
If an exabgp instance is configured with ip range as neighbor ip (passive neighbor), when a neighbor tries to connect the instance will fail with:

```
-- Traceback


Traceback (most recent call last):
  File "/home/wenxin/Projects/network/exabgp/lib/exabgp/application/bgp.py", line 426, in <module>
    main()
  File "/home/wenxin/Projects/network/exabgp/lib/exabgp/application/bgp.py", line 302, in main
    run(env,comment,configurations,root,options["--validate"])
  File "/home/wenxin/Projects/network/exabgp/lib/exabgp/application/bgp.py", line 370, in run
    exit_code = Reactor(configurations).run(validate,root)
  File "/home/wenxin/Projects/network/exabgp/lib/exabgp/reactor/loop.py", line 262, in run
    action = peer.run()
  File "/home/wenxin/Projects/network/exabgp/lib/exabgp/reactor/peer.py", line 622, in run
    if self.reactor.processes.broken(self.neighbor):
AttributeError: 'Listener' object has no attribute 'processes'

-- Logging History


20:30:04 | 10572  | cli control   |  - /home/wenxin/Projects/network/exabgp/var/run/0/
20:30:04 | 10572  | cli control   |  - /home/wenxin/Projects/network/exabgp/var/run/
20:30:04 | 10572  | cli control   | please make them in one of the folder with the following commands:
20:30:04 | 10572  | cli control   | > mkfifo /home/wenxin/Projects/NetLab/research/bgp-eam/examples/1on1/run/exabgp.{in,out}
20:30:04 | 10572  | cli control   | > chmod 600 /home/wenxin/Projects/NetLab/research/bgp-eam/examples/1on1/run/exabgp.{in,out}
20:30:04 | 10572  | configuration | performing reload of exabgp 4.0.10-a8462350
20:30:04 | 10572  | configuration | > neighbor         | '172.16.2.0/24'
20:30:04 | 10572  | configuration | . router-id        | '172.16.2.1'
20:30:04 | 10572  | configuration | . local-address    | '172.16.2.1'
20:30:04 | 10572  | configuration | . listen           | '179'
20:30:04 | 10572  | configuration | . local-as         | '65000'
20:30:04 | 10572  | configuration | . peer-as          | '65000'
[...]

PR #876, opened 2019-02-17
fix ranged neighbor Peer.reactor

wenxin-wang:
See https://github.com/Exa-Networks/exabgp/issues/875


wenxin-wang:
I found another bug with ranged neighbor (See #877), if the fix looks ok, should I include that fix in this pull request?

wenxin-wang:
I added fix for #877, and updated CHANGELOG. Hope it helps!

PR #879, opened 2019-03-07
Update parse-md5.conf

cliffburdick:
I believe it meant to say quotes instead of parentheses.


thomas-mangin:
Thank you cliff - this was indeed what was meant :-)

Issue #826, opened 2018-06-11
Extraneous address family added to peers after the first one configured.


asahi-eric:
##### Summary

4 peers configured, 2 of IPv4 only and 2 of IPv6 only, order is IPv4 peer, IPv6 peer, IPv4 peer, IPv6 peer.  Each peer is only configured for their respective address family but only the first peer configured in the config file adheres to that.   The remaining 3 try to use both address families.  Doing an `exabgpcli show neighbor configuration` shows that the 3 of the peers are configured for both address families.

##### OS
```
Linux Ubuntu 16.04.04 LTS
NAME="Ubuntu"
VERSION="16.04.4 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.4 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

```
Linux exabgp1 4.4.0-127-generic #153-Ubuntu SMP Sat May 19 10:58:46 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

##### Version
```
ExaBGP : 4.0.6-daa3d6ba
Python : 3.5.2 (default, Nov 23 2017, 16:37:01)  [GCC 5.4.0 20160609]
Uname  : Linux exabpg1 4.4.0-127-generic #153-Ubuntu SMP Sat May 19 10:58:46 UTC 2018 x86_64
Root   : /usr/share/exabgp1/exabgp
```

##### Installation
```
Custom fabric with pip3 install of whl
```

##### Environment 
```
empty
```

##### Configuration

Sanitized config.

```
process **** {
    run ****;
    encoder json;
}

neighbor 1.2.3.1 {
    router-id 1.2.3.4;
    local-address 1.2.3.4;
    local-as 64530;
    peer-as 64000;
    group-updates true;
    api {
        processes [ **** ];
        receive {
            parsed;
            update;
        }
        neighbor-changes;
    }
    family {
        ipv4 unicast;
    }
}
neighbor aaaa:bbbb:cccc:dddd::1 {
    router-id 1.2.3.4;
    local-address aaaa:bbbb:cccc:dddd::4;
    local-as 64530;
    peer-as 64000;
[...]

PR #881, opened 2019-03-20
Experimental fix for accumulating families

martintopholm:
When parsing multiple configuration address families from previous
neighbors appear to be added to subsequent peers (see issue #826).

This commit attempts to clear the parsed families after being used with
the current neighbor.


thomas-mangin:
Thank you Martin.

PR #886, opened 2019-05-01
Rfc5575bis review / treat as withdraw feature

stoffi92:
This pull request suggests the following changes to the code:

* General: Allow unpack_nlri of NLRI implementations to return a None value to MPRNLRI, MPURNLRI to allow a mechanism of "Treat as withdraw" handling of undecodable NLRI (this implementation is more a "ignore". Maybe the undecodeable NLRI should still be saved as a "Generic-unknown-NLRI" containing the undecoded byte-string received)
* Flowspec: Use the above to allow flowspec to treat undecodeable NLRI as withdraw according to RFC5575-bis
* Flowspec: Swap always "true"/"false" numeric operator (has been swapped during history the draft)
* Flowspec: TODO in flow.py line 551 -> why should the afi change during the livetime of the NLRI instance

I do not know if this is appropriate sending a pull-request this way (first pull-request ever). If you find time have a look

Cheers Christoph

### DIFF
--- a/lib/exabgp/bgp/message/update/nlri/flow.py
+++ b/lib/exabgp/bgp/message/update/nlri/flow.py
@@ -638,38 +638,41 @@ class Flow (NLRI):
 		bgp = bgp[:length]
 		nlri = cls(afi,safi,action)
 
-		if safi == SAFI.flow_vpn:
-			nlri.rd = RouteDistinguisher(bgp[:8])
-			bgp = bgp[8:]
+		try:
+			if safi == SAFI.flow_vpn:
+				nlri.rd = RouteDistinguisher(bgp[:8])
+				bgp = bgp[8:]
 
-		seen = []
+			seen = []
 
-		while bgp:
-			what,bgp = ordinal(bgp[0]),bgp[1:]
+			while bgp:
+				what,bgp = ordinal(bgp[0]),bgp[1:]
 
-			if what not in decode.get(afi,{}):
-				raise Notify(3,10,'unknown flowspec component received for address family %d' % what)
+				if what not in decode.get(afi,{}):
+					raise Notify(3,10,'unknown flowspec component received for address family %d' % what)
 
-			seen.append(what)
-			if sorted(seen) != seen:
-				raise Notify(3,10,'components are not sent in the right order %s' % seen)
+				seen.append(what)
+				if sorted(seen) != seen:
+					raise Notify(3,10,'components are not sent in the right order %s' % seen)
 
-			decoded = decode[afi][what]
-			klass = factory[afi][what]
+				decoded = decode[afi][what]
+				klass = factory[afi][what]
 
-			if decoded == 'prefix':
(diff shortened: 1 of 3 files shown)


lib/exabgp/bgp/message/update/nlri/flow.py
```diff
@@ -73,12 +73,12 @@ def length (data):
...
 	# reserved= 0x08  # 0b00001000
-	TRUE      = 0x00  # 0000000000
-	NEQ       = 0x06  # 0b00000110
 	LT        = 0x04  # 0b00000100
 	GT        = 0x02  # 0b00000010
 	EQ        = 0x01  # 0b00000001
-	FALSE     = 0x01 | 0x02 | 0x04
+	NEQ       = LT | GT
+	TRUE      = LT | GT | EQ
+	FALSE     = 0x00
```
thomas-mangin: This is quite a change with the current code and it inverts TRUE and FALSE ..

thomas-mangin:
How is the change of TRUE and FALSE going to affect existing users ? It could cause quite a surprise to current users of 4.0.x

thomas-mangin:
The reason for line 551 (as far as I can figure out myself). When we create a flow from the text parser, we do not know yet the family it is going to have.
```
@ParseFlow.register('route','append-route')
def route (tokeniser):
	change = Change(
		Flow(),
		Attributes()
	)
```

The code assume IPv4 .. but when it is then used, you will then specify `source-ipv6 ...` or `destination-ipv6` and we then change the family ... 

The issue with this is that it should also be done for all IPv6 only rules (and it looks like it is not done), so instead I should surely have a list of IPv4 and IPv4 rule and make sure that the flow is never inconsistent and change it for the others too.

stoffi92:
> How is the change of TRUE and FALSE going to affect existing users ? It could cause quite a surprise to current users of 4.0.x

This is arguable a change that may break things. However, those two cases (true, false) are more an academic completion of the possible bit combinations (was completely unspecified in RFC5575) and I do not think that this has any practical use. It is also going into the standard very soon. So at some point the change should go into the code as well (the original change was based on a very early version of the draft) 

If you think this TRUE/FALSE change should go into a later release I understand your concern and fine with it.

This is actually I am more interested in:
**"treat as withdraw" solution **
I think that this change has practical no impact of current users. This should reflect the behaviour described in rfc5575bis Section 11. But needs a change outside of the NLRI code.

Cheers Christoph

thomas-mangin:
4.0 is not yet stable so with a big fat warning we can perform this change ... It will need a BIG FAT warning tho (ie: an entry in the CHANGELOG + twitter + a wiki page for the 3->4 migration). I will do it.

stoffi92:
What do you think about the solution to allow an NLRI unpack method to return None to allow to treat undecodeable NLRI as withdraw? (Second and third commit).

thomas-mangin:
@stoffi92 merged :-) thanks a lot.

PR #890, opened 2019-05-22
Fix BGP MD5

AdamJacobMuller:
In cases where md5_bytes was already set (which seems to be almost always due to the creative base64 decoding), we were setting the `key` variable to null bytes. I'm not sure why that was done but it seems entirely wrong.

To fix that I adjusted things to always pack `md5_bytes` to `key`, but, only convert `md5` to `md5_bytes` if we don't already have `md5_bytes` set

### DIFF
--- a/lib/exabgp/reactor/network/tcp.py
+++ b/lib/exabgp/reactor/network/tcp.py
@@ -165,9 +165,8 @@ def MD5 (io, ip, port, md5, md5_base64):
 			if md5:
 				if md5_bytes is None:
 					md5_bytes = bytes_ascii(md5)
-					key = pack('2xH4x%ds' % TCP_MD5SIG_MAXKEYLEN, len(md5_bytes), md5_bytes)
-				else:
-					key = pack('2xH4x%ds' % TCP_MD5SIG_MAXKEYLEN, 0, b'')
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Thank you Adam for looking into it..
 
Is it not instead that the test should be inverted to "is not None" ?

thomas-mangin:
@AdamJacobMuller could you please let me know if the commit above fix the issue for you ? If so I will update the CHANGELOG :-)

AdamJacobMuller:
Hi,

I don't think that is the correct fix. shouldn't the logic here be "if we have md5 bytes, use that, if we don't have md5_bytes, convert md5 to md5_bytes and use that?"

thomas-mangin:
ack https://github.com/Exa-Networks/exabgp/commit/8ab6e14b9933fa47e6b3ac2f3deb0171fb13d9a0

thomas-mangin:
@AdamJacobMuller - I assume it is now working as should and I can close this PR ?

thomas-mangin:
no news == good news

AdamJacobMuller:
yep, looks good, thank you @thomas-mangin !

Issue #895, opened 2019-06-13
Healthcheck repeats announces


Marlinc:
Hey there,

We have noticed that our healthcheck checks keep re-announcing the prefix that they are testing over and over again. This without a change in the up/down status reported by the command run in `cmd`.

This only occurs when running the healthcheck as part of the ExaBGP daemon in a `process` section. Not when manually running the healthcheck (whether through root or the exabgp user). It doesn't appear to be a permission issue as the issue is also happening while using the exabgp user.

I have tested this on the version of ExaBGP as provided by Ubuntu 18.04 LTS 4.0.2-1c737d99 and the latest stable release 4.0.10

Here is some output when stracing one of the healthcheck processes, this appears to repeat every `--timeout` as expected. The only unexpected part is that it outputs the announce keyword every time it runs.

```
select(0, NULL, NULL, NULL, {tv_sec=1, tv_usec=889073}) = 0 (Timeout)
pipe2([5<pipe:[25555656]>, 6<pipe:[25555656]>], O_CLOEXEC) = 0
fstat(5<pipe:[25555656]>, {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
ioctl(5<pipe:[25555656]>, TCGETS, 0x7ffea9f19ae0) = -1 ENOTTY (Inappropriate ioctl for device)
lseek(5<pipe:[25555656]>, 0, SEEK_CUR)  = -1 ESPIPE (Illegal seek)
pipe2([7<pipe:[25555657]>, 8<pipe:[25555657]>], O_CLOEXEC) = 0
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7f427b7cfa10) = 14643
close(8<pipe:[25555657]>)               = 0
close(6<pipe:[25555656]>)               = 0
read(7<pipe:[25555657]>, "", 50000)     = 0
close(7<pipe:[25555657]>)               = 0
rt_sigaction(SIGALRM, {sa_handler=0x62ff10, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f427b203f20}, {sa_handler=0x62ff10, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f427b203f20}, 8) = 0
alarm(5)                                = 0
lseek(5<pipe:[25555656]>, 0, SEEK_CUR)  = -1 ESPIPE (Illegal seek)
[...]

PR #896, opened 2019-06-16
healthcheck: don't wait for acks when stdin is a terminal

vincentbernat:
When running interactively, users may be surprised to have to ack each
route like ExaBGP does. Therefore, when we detect stdin is a terminal,
don't wait for acks. Moreover, add an option to disable ack
processing (for people running with exabgp.api.ack set to False).

Related to #895.

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -38,6 +38,9 @@ match the beginning of the label without the interface prefix. In the
 example above, this means that you should have addresses on lo
 labelled ``lo:haproxy1``, ``lo:haproxy2``, etc.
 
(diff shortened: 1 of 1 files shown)


vincentbernat:
Oh, we both work on that in parallel!

Note you return while there are still more IPs to handle.

thomas-mangin:
Thanks for pointing my mistake, I will fix my code

thomas-mangin:
code fixed with https://github.com/Exa-Networks/exabgp/commit/ac47ed578d74b1f446fddca13d6751c73a9ccb39 sorry - I thought you expected me to write the patch ...

vincentbernat:
No problem!

PR #899, opened 2019-07-08
Add \n to CLI Open timeout error

cooperlees:
- It's annoying when debugging to not have the \n

### DIFF
--- a/lib/exabgp/application/cli.py
+++ b/lib/exabgp/application/cli.py
@@ -157,7 +157,7 @@ def main ():
 		sys.exit(0)
 
 	def open_timeout(signum, frame):
-		sys.stderr.write('could not open connection to ExaBGP')
+		sys.stderr.write('could not open connection to ExaBGP\n')
 		sys.stderr.flush()
 		sys.exit(1)
 


lib/exabgp/application/cli.py
```diff
@@ -157,7 +157,7 @@ def write_timeout(signum, frame):
 		sys.exit(0)
 
 	def open_timeout(signum, frame):
-		sys.stderr.write('could not open connection to ExaBGP')
+		sys.stderr.write('could not open connection to ExaBGP\n')
 		sys.stderr.flush()
```
cooperlees: I also feel Python shouldn't buffer stderr, but I could be wrong. So I left this for now.

cooperlees:
I didn't change any functional tests. Python 2 gives no error and just returns 1. Python 3 says:
```
  File "./qa/bin/functional", line 454, in <module>
    success = Command.dispatch(running)
  File "./qa/bin/functional", line 348, in dispatch
    if 'successful' in stdout:
TypeError: a bytes-like object is required, not 'str'
The command "./qa/bin/functional all" exited with 1.
```

Should that be if b'successful' in stdout or is stdout incorrect a b''? (haven't read the code)

thomas-mangin:
Thank you - https://github.com/Exa-Networks/exabgp/commit/af5d45e9455852e344cf4bbfdc00fcb9a0042082

Issue #900, opened 2019-07-09
Fix CI to Pass 


cooperlees:
I'd love to help with and see ✅ again on the CI.

TravisCI:
- Get passing again
- Add in an allowed failed job on Python 3.8 (to see if we have any incompatibilities) - e.g. https://github.com/pypa/bandersnatch/blob/master/.travis.yml

Scrutinizer:
- Possibly remove?

thomas-mangin:
Happy to remove Scrutinizer

thomas-mangin:
All the test fail on bionic - so I can try to reproduce

thomas-mangin:
Got it .. psutil seems to be python3 only (even if installed with pip install on bionic) and import fails even if pip was run before hand. 
Coverting the code to python3

thomas-mangin:
https://github.com/Exa-Networks/exabgp/commit/6ec1f15da093ad395527919c93444ec50b7bcdca

thomas-mangin:
work on bionic - works on Mac - does not work on travis - find the mistake !

thomas-mangin:
I have build a brand new Ubuntu machine following the steps in the CI files and it works.

thomas-mangin:
Two issues:
 - `pip install psutil` does not (always) work on the travis-ci VM ! for some reason
 - the docker image has no IPv6 loopback !

thomas-mangin:
Now now need to convert print format from python2 to python3 but unlocked

thomas-mangin:
Now passing

cooperlees:
Nice work! I’ll now have more confidence when changing things in the future.

cooperlees:
Can we please turn off the other failing CI? 🤞🏼 _(screenshot from head's CI runs)_
![Screen Shot 2019-07-16 at 1 38 05 PM](https://user-images.githubusercontent.com/3005596/61328031-0d04d480-a7cf-11e9-8d1a-e5e40af1ff39.png)
- continuous-integration/codeship
- Scrutinizer

I am sure these will be GUI clicking.

thomas-mangin:
I will try to get scrutiniser working to see if it is work keeping as in the past it found some bug doing linting.

PR #904, opened 2019-07-12
Remove scrutinizer

cooperlees:
- This does not work and we only intent to fix Travis
- No point running + reporting the fail

Part of #900


cooperlees:
This must also need some GUI clicking to completely stop Scrutinizer.

Issue #905, opened 2019-07-16
TravisCI not always running venv python


cooperlees:
I was trying to add Python 3.8 + nightly builds to ExaBGP CI, and found that `pypy` CI is not always actually running with Travis Environment's version as:
- `command -v python3` etc.

This lead to you needing the system apt installs. I'd like to remove (and will have a PR coming) to address this.

An example is the Python2 version was actually running with system Python 3.5:
```
$ sbin/exabgp --version
ExaBGP : (HEAD-b38529f31cb75097b72d125caf50d1a32f3526a4
Python : 3.5.2 (default, Nov 12 2018, 13:43:14)  [GCC 5.4.0 20160609]
Uname  : Linux travis-job-926aa7d8-30c3-41df-8372-4725b27a24a7 4.15.0-1028-gcp #29~16.04.1-Ubuntu SMP Tue Feb 12 16:31:10 UTC 2019 x86_64
Root   : /home/travis/build/Exa-Networks/exabgp
The command "sbin/exabgp --version" exited with 0.
```

thomas-mangin:
@cooperlees I assume we can not close this.

cooperlees:
Yeah, this is solved. Just need to see what bug PyPy and Python 2 are hitting.

thomas-mangin:
It looks like timing issue with the testing code ... sigh ! Not easy to change of fix.

cooperlees:
Why doesn’t it effect Python 3?

thomas-mangin:
it does affect all python. I had some times pypy fails and some other python2. It seems python3 behave a little bit better !

PR #906, opened 2019-07-16
[Travis] Test in more Python Versions + be more venv focused

cooperlees:
- Test in nightly and 3.8 builds + allow failure. Move to xenial to allow for this
- Move to matrix for fast finish
- Modify sbin/exabgp so that we will pickup venv python
- Ensure we're running latest pip, setuptools + wheel
- Install dependencies only into the venv - Remove apt installs ...
- Allow python2 + pypy to fail unitl we fix the functional tests

Addresses parts of #905

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,30 @@
 language: python
+dist: xenial
 
-python:
-  - "2.7"
-  - "3.7"
-  - "pypy"
-
-before_install:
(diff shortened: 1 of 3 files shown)


cooperlees:
Python 2 testing still broken, but I feel that will need more investigation and should be done on another pr.

PR #908, opened 2019-07-29
[healthcheck] Withdraw routes when exit

wavezhang:
The healthcheck  program remove ips when exit, but it donot withdraw routes associated. And packages still route to host, but there is no corresponding ips.  This patch withdraw routes when exit to  fix this bug.

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -465,7 +465,8 @@ def loop(options):
             setup_ips(options.ips, options.label, options.sudo)
 
         logger.info("send announces for %s state to ExaBGP", target)
-        metric = vars(options).get("{0}_metric".format(str(target).lower()), 0)
+        default_metric = vars(options).get("{0}_metric".format(str(states.DISABLED).lower()))
+        metric = vars(options).get("{0}_metric".format(str(target).lower()), default_metric)
         for ip in options.ips:
             if options.withdraw_on_down:
                 command = "announce" if target is states.UP else "withdraw"
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -349,14 +349,6 @@ def setup_ips(ips, label, sudo=False):
             subprocess.check_call(
                 cmd, stdout=fnull, stderr=fnull)
 
-    # If we setup IPs we should also remove them on SIGTERM
-    def sigterm_handler(signum, frame):  # pylint: disable=W0612,W0613
-        remove_ips(ips, label, sudo)
-        sys.exit(0)
-
-    signal.signal(signal.SIGTERM, sigterm_handler)
-
-
 def remove_ips(ips, label, sudo=False):
     """Remove added IP on loopback interface"""
     existing = set(loopback_ips(label))
@@ -456,14 +448,15 @@ def loop(options):
         "FALLING",              # Checks are currently failing.
         "UP",                   # Service is considered as up.
         "DOWN",                 # Service is considered as down.
+        "EXIT",                 # Exit state
     )
 
     def exabgp(target):
         """Communicate new state to ExaBGP"""
-        if target not in (states.UP, states.DOWN, states.DISABLED):
+        if target not in (states.UP, states.DOWN, states.DISABLED, states.EXIT):
             return
         # dynamic ip management. When the service fail, remove the loopback
-        if target in (states.DOWN, states.DISABLED) and options.ip_dynamic:
+        if target in (states.DOWN, states.DISABLED, states.EXIT) and options.ip_dynamic:
             logger.info("service down, deleting loopback ips")
             remove_ips(options.ips, options.label, options.sudo)
         # if ips was deleted with dyn ip, re-setup them
@@ -472,7 +465,7 @@ def loop(options):
             setup_ips(options.ips, options.label, options.sudo)
 
         logger.info("send announces for %s state to ExaBGP", target)
-        metric = vars(options).get("{0}_metric".format(str(target).lower()))
+        metric = vars(options).get("{0}_metric".format(str(target).lower()), 0)
         for ip in options.ips:
             if options.withdraw_on_down:
                 command = "announce" if target is states.UP else "withdraw"
@@ -601,6 +594,13 @@ def loop(options):
 
     checks = 0
     state = states.INIT
+    # Do cleanups on SIGTERM
+    def sigterm_handler(signum, frame):  # pylint: disable=W0612,W0613
+        exabgp(states.EXIT)
+        sys.exit(0)
+
+    signal.signal(signal.SIGTERM, sigterm_handler)
+
     while True:
         checks, state = one(checks, state)
 
--- a/lib/exabgp/reactor/api/processes.py
+++ b/lib/exabgp/reactor/api/processes.py
@@ -85,12 +85,12 @@ class Processes (object):
 			self._terminate(process)
 
 	def _terminate (self, process):
+		self.logger.debug('terminating process %s' % process,'process')
 		thread = Thread(target=self._terminate_run, args = (process,))
 		thread.start()
 		return thread
 
 	def _terminate_run (self, process):
-		self.logger.debug('terminating process %s' % process,'process')
 		try:
 			self._process[process].terminate()
 			self._process[process].wait()
--- a/lib/exabgp/reactor/api/processes.py
+++ b/lib/exabgp/reactor/api/processes.py
@@ -29,6 +29,7 @@ from exabgp.version import json as json_version
 from exabgp.version import text as text_version
 
 from exabgp.configuration.environment import environment
+from threading import Thread
 
 
 # pylint: disable=no-self-argument,not-callable,unused-argument,invalid-name
@@ -73,6 +74,8 @@ class Processes (object):
 		self._respawning = {}
 
 	def _handle_problem (self, process):
+		if process not in self._process:
+			return
 		if self.respawn_number:
 			self.logger.debug('issue with the process, restarting it','process')
 			self._terminate(process)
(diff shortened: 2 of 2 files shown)


vincentbernat:
Instead of `withdraw()`, it would be better to call `exabgp(states.EXIT)` and modify `exabgp()` function to handle this new state. This ensure some corner cases are correctly handled. For example, it will set `next-hop` to the correct value which is not always `self`. You could also remove the call to `remove_ips()` and reuse the one in `exabgp()`.

wavezhang:
@vincentbernat I've updated code, but the route cannot withdraw. stdout flush, but no bgp message send to peer, why？

lib/exabgp/application/healthcheck.py
```diff
@@ -456,14 +448,15 @@ def loop(options):
...
     )
 
     def exabgp(target):
         """Communicate new state to ExaBGP"""
-        if target not in (states.UP, states.DOWN, states.DISABLED):
+        if target not in (states.UP, states.DOWN, states.DISABLED, states.EXIT):
             return
         # dynamic ip management. When the service fail, remove the loopback
-        if target in (states.DOWN, states.DISABLED) and options.ip_dynamic:
+        if target in (states.DOWN, states.DISABLED, states.EXIT) and options.ip_dynamic:
```
vincentbernat: I think this is `target == states.EXIT or (target in (states.DOWN, states.DISABLED) and options.ip_dynamic):`

vincentbernat: My bad, your version is the right one.

lib/exabgp/application/healthcheck.py
```diff
@@ -472,7 +465,8 @@ def exabgp(target):
             setup_ips(options.ips, options.label, options.sudo)
 
         logger.info("send announces for %s state to ExaBGP", target)
-        metric = vars(options).get("{0}_metric".format(str(target).lower()))
+        if target != states.EXIT:
+            metric = vars(options).get("{0}_metric".format(str(target).lower()))
```
vincentbernat: I don't think this is needed. Mostly, `states.EXIT` is like `states.DOWN` with `options.withdraw_on_down` enabled. So, below, you should do `if options.withdraw_on_down or states.EXIT`.

lib/exabgp/application/healthcheck.py
```diff
@@ -507,7 +501,8 @@ def exabgp(target):
                         announce,
                         options.as_path)
 
-            metric += options.increase
+            if target != states.EXIT:
+                metric += options.increase
```
vincentbernat: Same here, not needed.

lib/exabgp/application/healthcheck.py
wavezhang: when exit and withdraw route， no need to change metric， but when announce， metric is needed

vincentbernat: This is handled below. As long as you change `if options.withdraw_on_down` to `if options.withdraw_on_down or target == states.EXIT`, `command` will be `withdraw` and metric and other stuff will not be used.

wavezhang: event options.withdraw_on_down is True, command still can be “announce” or “withdraw”， metric is still needed

vincentbernat: Below, if you do (just below):

```python
        for ip in options.ips:
            if options.withdraw_on_down or target == states.EXIT:
                command = "announce" if target is states.UP else "withdraw"
            else:
                command = "announce"
```

When you are in exit state, command will be `"withdraw"` and metric and other stuff is ignored. No need to add `if target == states.EXIT` elsewhere.

wavezhang: howerver the followings code will always be  executed, and will cause error since there is no EXIT_metirc
`metric = vars(options).get("{0}_metric".format(str(target).lower()))`
`metric += options.increase`

vincentbernat: You are right. I would suggest to use `vars(options).get("{0}_metric".format(str(target).lower()), 0)` to avoid this without litering the code with additional `if`.

wavezhang: code updated, but  routes still can not be withdrawed.

lib/exabgp/application/healthcheck.py
```diff
@@ -472,7 +465,7 @@ def exabgp(target):
             setup_ips(options.ips, options.label, options.sudo)
 
         logger.info("send announces for %s state to ExaBGP", target)
-        metric = vars(options).get("{0}_metric".format(str(target).lower()))
+        metric = vars(options).get("{0}_metric".format(str(target).lower()), 0)
         for ip in options.ips:
             if options.withdraw_on_down:
```
vincentbernat: You need to use `options.withdraw_on_down or target == states.EXIT`.

wavezhang: if withdraw_on_down is False, routes will not be withdrawed.

vincentbernat: That's why you need to *add* `or target == states.EXIT`. This way, the routes get withdrawn on exit whatever the value of `withdraw_on_down` is.

wavezhang: I mean when user specify withdraw_on_down to False, user donot want route to be withdrawed under any condition.

vincentbernat: Maybe not on down, but on exit, yes. Isn't the feature you want to implement? `withdraw_on_down` is `False` by default.

wavezhang: I just want route withdraw properly

vincentbernat: If I take your first commit and add this patch:

```diff
diff --git a/lib/exabgp/application/healthcheck.py b/lib/exabgp/application/healthcheck.py
index a11151c4887c..8631847646be 100644
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -467,7 +467,7 @@ def loop(options):
         logger.info("send announces for %s state to ExaBGP", target)
         metric = vars(options).get("{0}_metric".format(str(target).lower()), 0)
         for ip in options.ips:
-            if options.withdraw_on_down:
+            if options.withdraw_on_down or target is states.EXIT:
                 command = "announce" if target is states.UP else "withdraw"
             else:
                 command = "announce"
```

I am able to get a withdraw on exit:

```
INFO[healthcheck] send announces for UP state to ExaBGP
announce route 203.0.113.14/32 next-hop self med 100
INFO[healthcheck] send announces for UP state to ExaBGP
announce route 203.0.113.14/32 next-hop self med 100
INFO[healthcheck] send announces for EXIT state to ExaBGP
withdraw route 203.0.113.14/32 next-hop self
```

wavezhang: My test case：
1. initial config
```
process p1 {
	run python -m exabgp healthcheck  --ip IPADDR1 --withdraw-on-down --dynamic-ip-setup ;
	encoder text;
}

neighbor xxx {

	api {
		processes [ p1 ];
	}
}
```
2. change config to following：

```
process p2 {
	run python -m exabgp healthcheck  --ip IPADDR2 --withdraw-on-down --dynamic-ip-setup ;
	encoder text;
}

neighbor xxx {

	api {
		processes [ p2 ];
	}
}
```

3. reload config using exabgpcli
```
exabgpcli reload
```

wavezhang: My test case：
1. start exabgp with initial config
```
process p1 {
	run python -m exabgp healthcheck  --ip IPADDR1 --withdraw-on-down --dynamic-ip-setup ;
	encoder text;
}

neighbor xxx {

	api {
		processes [ p1 ];
	}
}
```
2. change config to following：

```
process p2 {
	run python -m exabgp healthcheck  --ip IPADDR2 --withdraw-on-down --dynamic-ip-setup ;
	encoder text;
}

neighbor xxx {

	api {
		processes [ p2 ];
	}
}
```

3. reload config using exabgpcli
```
exabgpcli reload
```

wavezhang: I've figure out that the cause is terminate process and bgp communication can not run parallelly. Terminate process need bgp communication to withdraw routes， but bgp communication code needs to wait termination done to continue。

wavezhang: I've fix this using new thread to terminate process, but there maybe better ways to do this.

thomas-mangin: Could you please explain why using thread is fixing the issue ?

wavezhang: run termination  in a new thread make terminate_process call return immediately， thus the program can go up to process  bgp communication.

lib/exabgp/reactor/api/processes.py
```diff
@@ -82,14 +85,19 @@ def _handle_problem (self, process):
 			self._terminate(process)
 
 	def _terminate (self, process):
+		thread = Thread(target=self._terminate_run, args = (process,))
+		thread.start()
+		return thread
+
+	def _terminate_run (self, process):
 		self.logger.debug('terminating process %s' % process,'process')
```
thomas-mangin: should the debug entry be moved to the `_terminate` function.

thomas-mangin:
merged and update the code
https://github.com/Exa-Networks/exabgp/commit/5c5af0456681c93ff442310bfc63ea6b17fd1bb3

thomas-mangin:
@vincentbernat I assume all if good from your POV ?

vincentbernat:
I still don't understand why we don't withdraw routes unconditionally on exit. We are asking a user to specify `--withdraw-on-down` when they want to withdraw on exit.

vincentbernat:
I would propose this patch:

```diff
diff --git a/lib/exabgp/application/healthcheck.py b/lib/exabgp/application/healthcheck.py
index a4462f2bd648..8631847646be 100644
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -465,10 +465,9 @@ def loop(options):
             setup_ips(options.ips, options.label, options.sudo)

         logger.info("send announces for %s state to ExaBGP", target)
-        default_metric = vars(options).get("{0}_metric".format(str(states.DISABLED).lower()))
-        metric = vars(options).get("{0}_metric".format(str(target).lower()), default_metric)
+        metric = vars(options).get("{0}_metric".format(str(target).lower()), 0)
         for ip in options.ips:
-            if options.withdraw_on_down:
+            if options.withdraw_on_down or target is states.EXIT:
                 command = "announce" if target is states.UP else "withdraw"
             else:
                 command = "announce"
```

thomas-mangin:
Thank you - merged :-)

wavezhang:
And thank you for your open-sourced projects :)

PR #920, opened 2019-09-25
update manpage to point to the correct redirect-to-ip draft

elindsey:
There are two drafts for flowspec redirect-to-ip. The manpage points
ietf, but simpson is the one currently implemented.

See eg. https://github.com/Exa-Networks/exabgp/blob/40f00b5e89e4a3f127639ac474b51371bd177b64/lib/exabgp/bgp/message/update/attribute/community/extended/traffic.py#L187


thomas-mangin:
Thank you

PR #921, opened 2019-09-26
Updating cli.py to stop hanging stdout

basyron:
When exabgp.api.ack environment config is set to "**false**", exabgpcli doesn't close exabgp.out fifo file (keeps inside an infinite loop (while not done)). This modification was made as a workaround to this situation.

Example environment:

```
[exabgp.api]
ack = false
chunk = 1
cli = true
compact = false
encoder = json
pipename = 'exabgp'
respawn = true
terminate = false

[exabgp.bgp]
openwait = 60

[exabgp.cache]
attributes = true
nexthops = true

[exabgp.daemon]
daemonize = false
drop = true
pid = ''
umask = '0o137'
user = 'nobody'

[exabgp.log]
all = false
configuration = true
daemon = true
destination = 'stdout'
enable = true
level = INFO
message = false
network = true
packets = false
parser = false
processes = true
reactor = true
rib = false
routes = false
short = false
timers = false

[exabgp.pdb]
enable = false

[exabgp.profile]
enable = false
file = ''

[exabgp.reactor]
speed = 1.0

[exabgp.tcp]
acl = false
bind = ''
delay = 0
once = false
port = 179
```

### DIFF
--- a/lib/exabgp/application/cli.py
+++ b/lib/exabgp/application/cli.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # encoding: utf-8
 """
 cli.py
--- a/lib/exabgp/application/cli.py
+++ b/lib/exabgp/application/cli.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # encoding: utf-8
 """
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Thank you @basyron - clearly I missed this case !

Issue #541, opened 2016-11-17
flowspec redirect ip implemention


pubyun:
some vendor like Huawei only implement new draft:

https://tools.ietf.org/html/draft-ietf-idr-flowspec-redirect-ip-02

and exabgp only implement old draft:

https://tools.ietf.org/html/draft-simpson-idr-flowspec-redirect-02

so the huawei and exabgp is not compatible

thomas-mangin:
Hello @pubyun 

Some other vendors also implement the old draft - so I am stuck and doomed to displease someone until the document reach RFC status ! :-(

Thomas

pubyun:
can ExaBGP support two draft and have a option for it?

thomas-mangin:
everything is possible but the realistic answer is not easily and I have no time to be able to start this work atm.

Johnny0326:
@thomas-mangin   Could you tell us where we can get the value of redirect ip of the config file in the code ? I want to try to make some changes to test the new draft 
[https://tools.ietf.org/html/draft-ietf-idr-flowspec-redirect-ip-02](url)
for example
	flow { 
		route test { 
			match { 
			destination 100.0.80.2/32; 
			#destination-port >8080&<8088 =3128; 
			source-port >1024; 
			protocol tcp; 
			} 
			then { 			
				**redirect 192.168.20.2;**
			} 
		      }
		}

thomas-mangin:
@Johnny0326 happy to help but .. I do not understand the question ..

Johnny0326:
I want to know where can I get value of the  parameter  of  redirect ip, then  I can assign it into the BGP Extended Communities.  In the draft https://tools.ietf.org/html/draft-simpson-idr-flowspec-redirect-02 the redirect IP is writed in the BGP Extended Communities.

thomas-mangin:
@Johnny0326 I still do not understand. The parser create a Change object which contains the NLRI and attribute the value will be within that object. You will need to read the code and understand the class structure if you want to perform changes.

The difference between the draft and the implementation change is very small, but as I said, as long as no RFC is out, I do not intend to change the code, unless all large vendors have already updated their implementation to the latest version.

sonixyznitesh:
@thomas-mangin (and others):

I know this thread is a year old but wanted to update with my observations as well:

[...]

PR #923, opened 2019-09-27
add support for draft-ietf-idr-flowspec-redirect-ip-02

elindsey:
Exabgp currently supports draft-simpson-idr-flowspec-redirect. This adds support for the newer draft of that standard, draft-ietf-idr-flowspec-redirect-ip.

Split across three commits to ease review - the first renames the current implementation so it doesn't conflict, the second adds ipv4 support, and the third is more involved and adds ipv6 support.

A few notes:
* Instead of having a flag (enable-ietf-redirect-mode) to change the interpretation of the current config syntax, I've added a new config directive (redirect-to-nexthop-ietf). This is because the next hop address needs to be encoded into the community itself whereas previously it was sent in a separate MP_REACH_NLRI. To preserve the existing syntax would require state in the parser (recording the next-hop directive for later use).
* I've included the copy bit, but haven't bothered exposing it via config. I don't think it's very common, and I'm not sure if it's worth cluttering up with another copy-ietf directive.
* Static parser - I can't tell exactly what this component does or if I need to add support for this draft there as well. Any pointers?
* Please let me know if there are any name changes you'd like. This is a bit messier than I had hoped, but it's also a messy situation (duplication across two drafts and IPv4/IPv6)

Testing:
I've run this locally against a Juniper vMX to verify that routes are received correctly. I couldn't tell if there were existing unit tests for the old redirect draft implementation - please let me know if there's somewhere I should add tests and how to trigger them.

Test config
```
neighbor 172.16.19.2 {
    router-id 172.16.19.1;
    local-address 172.16.19.1;
    local-as 100;
    peer-as 100;
    flow {
        route testflowipv4 {                                                                                                                
            match {
                source 192.0.2.42/32;
            } then {
               redirect-to-nexthop-ietf 198.51.100.2;
            }   
        }   
        route testflowipv6 {
            match {
                source fd00:0000:0000::1/128;
            } then {
               redirect-to-nexthop-ietf FE80:0000:0000::2;
            }   
        }   
    }   
}
```
Junos output (ipv6 path is ficticious because I didn't have a route to the address in my test environment, I was just making sure I could receive a route)
```
root> show route table inet6flow.0           

inet6flow.0: 1 destinations, 1 routes (1 active, 0 holddown, 0 hidden)
+ = Active Route, - = Last Active, * = Both

*,fd00::1/128/term:1            
                   *[BGP/170] 00:00:07, localpref 100, from 172.16.19.1
                      AS path: I, validation-state: unverified
                       Fictitious

root> show route table inetflow.0 

inetflow.0: 1 destinations, 1 routes (1 active, 0 holddown, 0 hidden)
+ = Active Route, - = Last Active, * = Both

*,192.0.2.42/term:1            
                   *[BGP/170] 00:00:12, localpref 100, from 172.16.19.1
                      AS path: I, validation-state: unverified
                    >  to 198.51.100.2 via ge-0/0/9.0
```

Related to https://github.com/Exa-Networks/exabgp/issues/541


thomas-mangin:
The patches look good, perhaps some space vs tabs here and there and a need to add some QA code for the new encoding but nothing which would prevent merging.

elindsey:
Thanks Thomas! 

I looked over everything again for for tabs/spaces issues but might be missing it - it looked like the style was tabs for indentation, spaces for internal alignment, so I tried to follow that.

I added a QA test for the new redirect-to-nexthop-ietf config. 

One last note about testing: I've validated the packet format manually, confirmed that arista accepts both ipv4 and ipv6 next hops, and confirmed that junos accepts ipv4 next hops. However junos doesn't like the ipv6 next hop attributes - still trying to figure out why, but looking like a vendor issue.

thomas-mangin:
Thank you for all this - just missing the credit entry :-) I will do that for you !

thomas-mangin:
If you find the reason for the IPv6 bug please open another issue and I will merge. It would not be the first time a vendor FlowSpec implementation would need a review.

elindsey:
@thomas-mangin I confirmed with Juniper that they don't currently support ipv6 as a next hop for draft-ietf-idr-flowspec-redirect-ip-02, so everything's as expected.

thomas-mangin:
need perfect patch - the registration code needed to be made generic and the check use the right sub-class :-)

Oleksa31:
>>>I've included the copy bit, but haven't bothered exposing it via config. I don't think it's very common, and I'm not sure if it's worth cluttering up with another copy-ietf directive.>>>
Hello guys,
Could you fully enable this functionality in exabgp? It will be very useful for network troubleshooting.
Thanks.

PR #928, opened 2019-10-21
fix bugs in remove_ips

wavezhang:
Fix logic errors when remove_ips


thomas-mangin:
@vincentbernat patch seems right - merging.

Issue #930, opened 2019-10-23
restart via 


juise:
##### Summary

call `restart` via tcp api...

##### OS

```
Linux ...  4.18.0-041800-generic #201808122131 SMP Sun Aug 12 21:33:20 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

##### Version

```
exabgp --version
ExaBGP : 4.1.2-d006a34a
Python : 2.7.12 (default, Aug 22 2019, 16:36:40)  [GCC 5.4.0 20160609]
Uname  : Linux ... 4.18.0-041800-generic #201808122131 SMP Sun Aug 12 21:33:20 UTC 2018 x86_64
Root   : /usr/local
```

##### Installation

```
pip
```

##### Environment 

```
# cat /etc/exabgp/exabgp.env
[exabgp.daemon]
# should we run in the background.                                          default (false)
daemonize = true
# where to save the pid if we manage it.                                    default ('')
pid = ''
# user to run as.                                                           default ('nobody')
user = 'root'
# drop privileges before forking processes.                                 default (true)
drop = true

[exabgp.log]
# enable logging.                                                           default (true)
enable = true
# log message with at least the priority SYSLOG.<level>.                    default (INFO)
#level = INFO
# where logging should log...                                               default ('stdout')
#destination = 'host:127.0.0.1'

# report changes in route announcement on config reload.                    default (false)
message = true
# report networking information (TCP/IP, network state,...).                default (true)
network = true

# all = true
packets = true
parser = true
processes = true
reactor = true
rib = true
routes = true
timers = true

[exabgp.bgp]
# how many second we wait for an open once the TCP session is established.  default (60)
openwait = 60
```

##### Configuration

```
process tcp-control-server {
    run ./example-tcp-control.run;
    encoder text;
}

template {
    neighbor common {
        router-id 127.0.0.1;
        local-address 127.0.0.1;

        local-as 65533;
        peer-as 123;

        api services {
            processes [ tcp-control-server ];
        }
    }
}

neighbor 1.1.1.1 { inherit common; }
neighbor 1.1.1.2 { inherit common; }
```

##### Program output

```
13:54:37 | 23218  | daemon        | ExaBGP can not fork when logs are going to stdout
13:54:37 | 23218  | configuration | performing reload of exabgp 4.1.2-d006a34a
13:54:37 | 23218  | reactor       | loaded new configuration successfully
[...]

PR #931, opened 2019-10-23
fix issue #930

juise:


thomas-mangin:
Thank you for the fix .. which would work but it would remove the message. We should get the neighbour from the peer instead - Could I ask you to do so if you do not mind ?
Also add yourself to the CHANGELOG for the work please :-)

thomas-mangin:
If not I will do it a bit later on.

juise:
Yeah, done!

lib/exabgp/configuration/neighbor/api.py
```diff
@@ -139,7 +139,7 @@ def _empty (cls):
 	def clear (self):
 		self.api = {}
 		self.named = ''
-		pass
+		Section.clear(self)
```
thomas-mangin: Did you find anything wrong that this fixes please ?

juise: Yeah, on 'restart', the follow code https://github.com/Exa-Networks/exabgp/blob/2c27112b480132be12a1b3290704f0312769eef9/lib/exabgp/configuration/neighbor/api.py#L147 returns exception/error like

```
19:31:13 | 1342   | configuration | not reloaded, no change found in the configuration
19:31:13 | 1342   | configuration | problem parsing configuration file line 21
19:31:13 | 1342   | configuration | error message: the name "services" already exists in api
```

juise: Could you please, create the new build 4.1.3 and upload it into the PyPI? )

PR #933, opened 2019-11-08
Don't resolve configuration symlinks

juise:


thomas-mangin:
Hello @juise - thank you for the PR but why should we not do this in your opinion ?

juise:
I’m using symlinks for atomic configuration change, and to keep history. So, every changes in configuration written in the new file, and symlink to it atomically changed, after that I call “reload” command. But... config link is resolved into first symlinked file by exabgp... So, I think it will be better don’t resolve symlinks explicitly

thomas-mangin:
I understand the use case. However, the test we are performing should be against the file and not the symlink, and we do not want to have a race condition with someone changing the symlink between the test and opening the file (as ExaBGP can run as root). It is not as bad as we do not write but still good practice.

thomas-mangin:
@juise - thank you implemented differently with https://github.com/Exa-Networks/exabgp/commit/594000bc382afe359b63a1b69a97b597a38d262a but the result should be the same.

juise:
Awesome, this fix very useful for me, thanks!

PR #939, opened 2019-12-09
Move import of lsprofcalltree in bgp.py

cooperlees:
- This needs to be moved to only be imported IF profiling is enabled in the env file

Test: 
Applied this patch locally and saw `exabgp` start when there is no cProfile module available

### DIFF
--- a/lib/exabgp/application/bgp.py
+++ b/lib/exabgp/application/bgp.py
@@ -22,7 +22,6 @@ from exabgp.version import version
 from exabgp.reactor.loop import Reactor
 
 from exabgp.vendoring import docopt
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Hello  @cooperlees - why is this required ? As the module is vendor, the file is there anyway and the import should not be affecting performance and/or runtime.

cooperlees:
The `cProfile` module is not vendored:
- https://github.com/Exa-Networks/exabgp/blob/master/lib/exabgp/vendoring/lsprofcalltree.py#L13

In smaller distributed Linux builds uneeded core parts of CPython are left out. We leave out cProfile. Move the import allows us to function without needed this library.

Issue #788, opened 2018-02-21
What should a normal exabgp return code be with systemctl stop?


martbhell:
##### ISSUE TYPE

 - Bug Report

##### OS

 - CentOS 7.4
 - python2

##### VERSION

```
# exabgp --version
ExaBGP : 3.4.23
Python : 2.7.5 (default, Aug  4 2017, 00:39:18)  [GCC 4.8.5 20150623 (Red Hat 4.8.5-16)]
Uname  : #1 SMP Thu Jan 4 01:06:37 UTC 2018
```

##### ENVIRONMENT 

```
# exabgp --di

[exabgp.log]
short = true

```

##### SUMMARY

When exabgp service is stopped with systemd then it returns return code 1.

Is return code 1 a normal exit? If it is I can make a PR to update the systemd service file :)

Setting SuccessExitStatus in the systemd service file makes systemd status happy:
 
<pre>
[Service]
SuccessExitStatus=1
</pre>
##### STEPS TO REPRODUCE

systemctl stop exabgp

##### EXPECTED RESULTS

systemctl status says the system OK, not degraded because of a failed service (exabgp)

<pre>
# systemctl status ceph-radosgw@obfusc.service
● ceph-radosgw@obfusc.service - Ceph rados gateway
   Loaded: loaded (/usr/lib/systemd/system/ceph-radosgw@.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Wed 2018-02-21 14:23:46 EET; 8min ago
  Process: 14006 ExecStart=/usr/bin/radosgw -f --cluster ${CLUSTER} --name client.%i --setuser ceph --setgroup ceph (code=exited, status=0/SUCCESS)
 Main PID: 14006 (code=exited, status=0/SUCCESS)
</pre>

##### ACTUAL RESULTS

<pre>
# systemctl status exabgp
● exabgp.service - ExaBGP
   Loaded: loaded (/etc/systemd/system/exabgp.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2018-02-21 14:31:26 EET; 1s ago
     Docs: https://github.com/Exa-Networks/exabgp/wiki
 Main PID: 16938 (exabgp)
   CGroup: /system.slice/exabgp.service
           ├─16938 /usr/bin/python2 /bin/exabgp /etc/exabgp/exabgp.conf
           └─16941 /usr/bin/python -m exabgp healthcheck --config /etc/exabgp/healthcheck_ipv4.conf

[...]

PR #943, opened 2020-01-09
Return 0 on exit

badrabubker:
#788


thomas-mangin:
Thank you. well spotted !

Issue #951, opened 2020-02-04
loopback addresses not detected if no labels in use


ruben-herold:
##### Summary

loopback addresses not detected if no labels in use

##### OS

centos 8 

bash-4.4# uname -a
Linux box 4.18.0-147.3.1.el8_1.x86_64 #1 SMP Fri Jan 3 23:55:26 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
bash-4.4# 

##### Version
```
bash-4.4# /usr/local/bin/exabgp --version
ExaBGP : 4.2.4
Python : 3.6.8 (default, Nov 21 2019, 19:31:34)  [GCC 8.3.1 20190507 (Red Hat 8.3.1-4)]
Uname  : Linux box 4.18.0-147.3.1.el8_1.x86_64 #1 SMP Fri Jan 3 23:55:26 UTC 2020 x86_64
Root   : /usr/local
bash-4.4# 

```

##### Installation
```
pip3 install exabgp

```

The loopback addresses are not detected if no lables in use:

```
bash-4.4# ip addr 
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet XXX.XXX.XXX.XXX/32 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever

bash-4.4# python3 -m exabgp healthcheck -d --name jool --no-ip-setup --start-ip 1  --interval 10 --cmd /bin/true
DEBUG[healthcheck] Retrieve loopback IP addresses
DEBUG[healthcheck] Loopback addresses: []
ERROR[healthcheck] No IP found
```

Fix is very simple will add Pull request

thomas-mangin:
@ruben-herold I assume this issue can be closed.

thomas-mangin:
@vincentbernat I may have been a bit quick to release:
https://github.com/Exa-Networks/exabgp/commit/4c9463c29ea1ab446be48a91f510252243dfd165

could you please review ?

thomas-mangin:
Thank you. updated and released 4.2.6

ruben-herold:
Runs fine thx for the merge

PR #952, opened 2020-02-04
Fix loopback detection without label issue #951

ruben-herold:
Fix loopback detection without label issue #951


thomas-mangin:
Thank you @ruben-herold

Issue #967, opened 2020-04-17
Remove path attributes in an UPDATE message that only contains MP_UNREACH_NLRI


subsecond:
##### Summary
Following the principle "be liberal in what you accept and conservative in what you send" we should remove the path attributes `ORIGIN` and `AS_PATH` from UPDATE messages that only contain the `MP_UNREACH_NLRI` because [RFC4760](https://tools.ietf.org/html/rfc4760) states:

>    An UPDATE message that contains the MP_UNREACH_NLRI is not required
>    to carry any other path attributes.

Currently, when sending a BGP `withdraw`, here is what that update message looks like on the wire:

```
21:12:32.103664 fa:16:3e:22:0a:8b > fa:16:3e:4f:6f:0b, ethertype IPv6 (0x86dd), length 153: (flowlabel 0x491a8, hlim 64, next-header TCP (6) payload length: 99) 2001:db8::200.41291 > 2001:db8::100.179: Flags [P.], cksum 0xc11d (correct), seq 77:136, ack 1, win 507, options [nop,nop,md5 valid], length 59: BGP
	Update Message (2), length: 59
	  Origin (1), length: 1, Flags [T]: IGP
	    0x0000:  00
	  AS Path (2), length: 6, Flags [T]: 65200 
	    0x0000:  0201 0000 feb0
	  Multi-Protocol Unreach NLRI (15), length: 20, Flags [O]: 
	    AFI: IPv6 (2), SAFI: Unicast (1)
	      2001:db8:dead:beef::1/128
	    0x0000:  0002 0180 2001 0db8 dead beef 0000 0000
	    0x0010:  0000 0001
```

Both `ORIGIN` and `AS_PATH` are path attributes that, according to [RFC4760](https://tools.ietf.org/html/rfc4760), are not required to be sent in the update message (I am fully aware that this is not a "must not" but more of a "should not").

##### OS
```
Debian Linux
```

```
root@exabgp-test:~# uname -a
Linux exabgp-test 4.19.0-8-cloud-amd64 #1 SMP Debian 4.19.98-1 (2020-01-26) x86_64 GNU/Linux
```

##### Version
```
root@exabgp-test:~# exabgp --version
ExaBGP : 4.2.6
Python : 2.7.16 (default, Oct 10 2019, 22:02:15)  [GCC 8.3.0]
[...]

PR #968, opened 2020-04-22
Exclude path attributes from MP_UNREACH_NLRI only messages

href:
See Exa-Networks/exabgp#967.

I implemented this knowing little about ExaBGP's code base, so I'm not sure this suffices. I did verify that all tests still succeed, so there's at least that.

I also did try to add a test for this specific case, but I didn't quite understand how to best approach this and after some tries I gave up.

We did run this change in our lab and in two separate developer environments, observing the effects, testing general functionality, and looking at the messages on the wire. We observed the desired effects and it solves our problem.

As this issue is of some importance to us, because it solves a real issue we face, we would appreciate it a lot if this PR could be reviewed, merged, and released swiftly. Though it has to be said that this is not a bug in ExaBGP. It would simply be really nice 🙂.

Let me know how we can support you, and do get back to me if you want me to change or add something.

### DIFF
--- a/lib/exabgp/bgp/message/update/__init__.py
+++ b/lib/exabgp/bgp/message/update/__init__.py
@@ -122,7 +122,28 @@ class Update (Message):
 		if not nlris and not mp_nlris:
 			return
 
-		attr = self.attributes.pack(negotiated, True)
+		# If all we have is MP_UNREACH_NLRI, we do not need the default
+		# attributes. See RFC4760 that states the following:
+		#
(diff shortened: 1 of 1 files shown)


href:
Since the master branch seems to encounter the same errors as this PR, and because the same tests work locally, I shall assume that the failed Travis tests are not my doing.

thomas-mangin:
Sorry, due to the use of black of the patch could could not be merged as it. Manually merged it. Thank you.

PR #977, opened 2020-05-17
Improve systemd unit configs

nnvai47:
There are two changes here:

* set user and group to 'exabgp'. Without this I can't use
  'exabgpcli'. For example, without this 'sudo exabgpcli version'
  fails with "could not send command to ExaBGP (command timeout)".

* use absolute paths in the 'ExecStartPre' commands. This doesn't seem
  to matter on Debian 10, but on Ubuntu 18.04 the non-absolute-path
  versions cause an error.

Note that if you're testing these changes by editing
/lib/systemd/system/exabgp.service, you need to run

    systemctl daemon-reload

before

    systemcl restart exabgp

for the changes to take effect.

### DIFF
--- a/etc/systemd/exabgp.service
+++ b/etc/systemd/exabgp.service
@@ -7,16 +7,16 @@ After=network.target
 ConditionPathExists=/etc/exabgp/exabgp.conf
 
 [Service]
-#User=exabgp
-#Group=exabgp
+User=exabgp
+Group=exabgp
 Environment=exabgp_daemon_daemonize=false
 PermissionsStartOnly=true
-ExecStartPre=-mkfifo /run/exabgp.in
-ExecStartPre=-mkfifo /run/exabgp.out
-ExecStartPre=chmod 600 /run/exabgp.in
-ExecStartPre=chmod 600 /run/exabgp.out
-ExecStartPre=chown exabgp.exabgp /run/exabgp.in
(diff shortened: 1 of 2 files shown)


thomas-mangin:
sure will merge - thank you for the contribution.

nnvai47:
Could someone (@vincentbernat ?) cherry pick this into the `debian/sid` branch?

vincentbernat:
It will get there automatically at the next release.

nnvai47:
OK, great! Sorry to waste your time :)

nnvai47:
Hello again, it appears this never made it into the Debian Sid branch, e.g. here's the latest Debian Sid version of the exabgp.service file: https://github.com/Exa-Networks/exabgp/blob/debian/sid/4.2.8-2/etc/systemd/exabgp.service .

vincentbernat:
Oh, I am sorry. I have totally forgot about that and got pinged again only this week-end. 4.2.8-2 got a patch (https://github.com/Exa-Networks/exabgp/blob/debian/sid/4.2.8-2/debian/patches/0001-Patch-exabgp.service-to-fix-socket-creation-on-start.patch) which is mostly a simplification of your changes by putting the socket into a directory. As Debian is freezed, I cannot push a whole new version in debian/sid.

nnvai47:
Sounds good, thanks for the update.

PR #981, opened 2020-05-26
issue966:keep old process's stdio pipes until healthcheck processes r…

shaoberry:
…eceive ack and exit


thomas-mangin:
Thank you very much for the patch, I will review it and see if it is ok.

Please keep in mind that not all users for ExaBGP use it with the healthchecker. Some have some code using their OWN DB, etc and may not expect two of them to be able to run at the same time.

thomas-mangin:
It seems to break the QA code tho so clearly can not be merged as it 😉

PR #988, opened 2020-07-23
Migrating from master and slave phrasing

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.


thomas-mangin:
Please get the vendors to change the name of their software and then the reference can be updated.

I am all for diversity, respect, and fairness in life for all, but I do not believe this patch is where the effort should be spent. I appreciate your sentiment though.

PR #994, opened 2020-10-12
healthcheck: remove loopback ips after sending BGP announce

### LINKED ISSUE #993: Healthcheck with dynamic-ip-setup remove loopback ips before announcing DOWN/DISABLED
**The bug**

When the healthcheck is failing the loopback ips are removed prior to announce the DOWN (or DISABLED), resulting in a short period during which packets routed to the the server are refused.

>python[18086]: healthcheck[18086]: Check command was unsuccessful: 1
python[18086]: healthcheck[18086]: service down, deleting loopback ips
sudo[18156]:   exabgp : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/usr/sbin/ip address delete 192.0.0.1/32 dev lo
python[18086]: healthcheck[18086]: send announces for DISABLED state to ExaBGP
exabgp[18077]: 12:48:29 | 18077  | api | route removed from neighbor 192.0.0.254 local-ip 10.0.0.137 local-as 64520 peer-as 65001 router-id 10.0.0.137 family-allowed in-open : 192.0.0.1/32 next-hop self

**To Reproduce**

1. setup healthcheck on the vip
```
process watch-vip {
  run /bin/python -m exabgp healthcheck --name 'Check vip' --command  '/opt/health_check.sh' --ip 192.0.0.1 --sudo --dynamic-ip-setup --disable /run/pdns_disable --withdraw-on-down;
  encoder 'text';
}
```
2. kill the service or create the disable file `sudo touch /run/pdns_disable `

**Expected behavior**

The service might still be a bit working even if the healthcheck is starting to fail, or has the disable file created.
Without the loopback ips, it sure it won't work anymore as the packet are going to be refused, so I expect the healthcheck script to first announce the DOWN (or DISABLED) state then remove the loopback ips.

The log output should be something like that:
> python[2001]: healthcheck-Internal DNS zones[2001]: send announces for DISABLED state to ExaBGP
exabgp[1993]: 13:47:15 | 1993   | api | route removed from neighbor 192.0.0.254 local-ip 10.0.0.137 local-as 64520 peer-as 65001 router-id 10.0.0.137 family-allowed in-open : 192.0.0.1/32 next-hop self
python[2001]: healthcheck-Internal DNS zones[2001]: service down, deleting loopback ips
sudo[7990]:   exabgp : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/usr/sbin/ip address delete 192.0.0.1/32 dev lo

**Environment (please complete the following information):**

 - OS: CentOS 7.8
 - Version 4.1.2 (custom)

Annih:
Otherwise some packets can still be routed to the server which is going to refuse them.

Fix #993

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -358,14 +358,6 @@ def loop(options):
             return
         if target in (states.END,):
             return
-        # dynamic ip management. When the service fail, remove the loopback
-        if target in (states.EXIT,) and (options.ip_dynamic or options.ip_setup):
-            logger.info("exiting, deleting loopback ips")
-            remove_ips(options.ips, options.label, options.sudo)
-        # dynamic ip management. When the service fail, remove the loopback
-        if target in (states.DOWN, states.DISABLED) and options.ip_dynamic:
-            logger.info("service down, deleting loopback ips")
-            remove_ips(options.ips, options.label, options.sudo)
         # if ips was deleted with dyn ip, re-setup them
         if target == states.UP and options.ip_dynamic:
             logger.info("service up, restoring loopback ips")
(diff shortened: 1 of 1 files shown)


vincentbernat:
Ack for me.

thomas-mangin:
Thank you very much @Annih - it would be great if you could provide the same PR for the 4.2 branch as master is in flux (the command line changed, and more is to come) and this should clearly be backported.

Annih:
Done in #995  @thomas-mangin
Thank you for the quick reaction!

PR #995, opened 2020-10-13
healthcheck: remove loopback ips after sending BGP announce

Annih:
Otherwise some packets can still be routed to the server which is going to refuse them.


thomas-mangin:
Brilliant, thank you Baptist - I just must not forget to update both CHANGELOG to include your contribution :-)

Issue #938, opened 2019-11-29
Manpages are outdated


jsenecal:
##### Summary

ExaBGP [manpage](https://github.com/Exa-Networks/exabgp/blob/master/doc/man/exabgp.1) needs to reflect the proper (current) parameters.

For instance, `--folder` is no longer a supported parameter.

Thanks,

##### OS

```
Linux E7470-JS 5.3.13-arch1-1 #1 SMP PREEMPT Sun, 24 Nov 2019 10:15:50 +0000 x86_64 Linux
```

##### Version

```
ExaBGP : 4.1.2-d006a34a
Python : 3.7.4 (default, Aug 21 2019, 00:19:59)  [GCC 8.3.0]
Uname  : Linux E7470-JS 5.3.13-arch1-1 #1 SMP PREEMPT Sun, 24 Nov 2019 10:15:50 +0000 x86_64
Root   : /usr/local
```

##### Installation

Installing from pip, but the [versions available in Debian testing/unstable](https://dyn.manpages.debian.org/testing/exabgp/exabgp.1.en.html) are the same
```
pip install --upgrade exabgp==4.1.2
```

##### Environment 

```
[exabgp.daemon]
drop = false
```

##### Program output

```
# /usr/local/bin/exabgp --env /etc/exabgp/exabgp.env --folder /etc/exabgp/
usage: exabgp [--help] [--version]
              [--root ROOT] [--env ENV]
              [[--full-ini | --diff-ini | --full-env | --diff-env] |
              [--fi | --di | --fe | --de]]
              [--debug] [--pdb] [--test]
              [--once] [--signal TIME]
              [--memory] [--profile PROFILE]
              [--validate]
              [--run HELPER]
              [--decode HEX_MESSAGE]...
              [<configuration>...] 
```
##### Importance

Documentation issue (manpage)

thomas-mangin:
Thank you for reporting the issue. As I did not write the initial man page, any help would be appreciated.

jkldgoefgkljefogeg:
is `-e` deprecated? env file path seems to ignore -e
https://github.com/Exa-Networks/exabgp/blob/master/src/exabgp/environment/base.py#L4,L8

thomas-mangin:
Correct on master, to be version 5, the env file is taken from the configuration folder.

thomas-mangin:
Man pages from the v4 era. ExaBGP 5.x has new CLI with built-in help (`help` command). Closing as outdated.

PR #1000, opened 2020-11-07
Fix the man page

MrRio:
This relates to issue #938 

- Change the order of arguments to match the script output
- Remove folder arg
- Add and document root, run, validate arguments

[exabgp.1.pdf](https://github.com/Exa-Networks/exabgp/files/5504638/exabgp.1.pdf)

We're coming in to see Mark on Friday so thought we'd have a look, nice tool!

### DIFF
--- a/doc/man/exabgp.1
+++ b/doc/man/exabgp.1
@@ -6,13 +6,17 @@
 .Nd Influence or control network using BGP
 .Sh SYNOPSIS
 .Nm
-.Op Fl -folder Ar folder | Fl f Ar folder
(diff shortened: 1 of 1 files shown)


thomas-mangin:
The command line on master changed and is not the same as for 4.x. So applying this patch to master does not make sense.
Could you please port the patch to the 4.2 branch so I can apply it? It would be much appreciated. Thank you.

MrRio:
@thomas-mangin Yep - no problem!

MrRio:
New PR submitted in #1001

PR #1001, opened 2020-11-08
4.2 Fix the man page and text tweaks

MrRio:
Updated to the 4.2 branch. This relates to issue #938 

- Change the order of arguments to match the script output
- Remove folder arg
- Add and document root, run, validate arguments
- Fix grammar on openwait environment
- Change exabgp to ExaBGP in usage help for consistency with the line below

[exabgp.1.pdf](https://github.com/Exa-Networks/exabgp/files/5504638/exabgp.1.pdf)

### DIFF
--- a/doc/man/exabgp.1
+++ b/doc/man/exabgp.1
@@ -6,13 +6,17 @@
 .Nd Influence or control network using BGP
 .Sh SYNOPSIS
 .Nm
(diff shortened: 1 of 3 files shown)


thomas-mangin:
Thank you very much James. Have a very good day!

PR #1008, opened 2020-11-30
Add --*-as-path options to change as-path on healthcheck status. #1007

### LINKED ISSUE #1007: Healthcheck: Changing/extending AS path on up/down
https://github.com/Exa-Networks/exabgp/blob/933c759a5dde0f44b7d9ff13f8fdaa00be04b35a/lib/exabgp/application/healthcheck.py#L367

Is it possible to add some new flags like `--up-as-path` and `--down-as-path`? This would be the same as the `--*-metric` commands but affecting the `--as-path` flag. Some providers don't support ECMP based on metric but rather the only way to withdraw is to double up the path so it is not the preferred one any more?

mzealey:
Fixes #1007

### DIFF
--- a/lib/exabgp/application/healthcheck.py
+++ b/lib/exabgp/application/healthcheck.py
@@ -213,12 +213,14 @@ def parse():
         help="announce IPs with the supplied community when disabled",
     )
     g.add_argument("--as-path", metavar="ASPATH", type=str, default=None, help="announce IPs with the supplied as-path")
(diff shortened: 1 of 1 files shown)


vincentbernat:
Looks good for me.

thomas-mangin:
@mzealey could you please forward port to master too?

mzealey:
@thomas-mangin this was actually already on mzealey:master branch if you want to pull it over

Issue #1006, opened 2020-11-26
USR2 starts new healthcheck process but doesn't terminate the current one


mzealey:
**Describe the bug**

When I `kill -USR2` exabgp the config reloads and a new healthcheck process is started, but the old ones stick around.

**To Reproduce**

I have a process set up like:

```
process healthcheck {
        run python3 -m exabgp healthcheck --no-syslog --ip xxx --next-hop xxx --cmd "true" --no-ip-setup --community "xxx" --up-metric 400 --withdraw-on-down;
        encoder text;
}
```

I see the following debug output when sending reload signal to exabgp:

```
18:17:39 | 1      | reactor       | SIGUSR2 received
18:17:39 | 1      | reactor       | scheduling reload of configuration and processes
18:17:39 | 1      | configuration | performing reload of exabgp 4.2.11
18:17:39 | 1      | configuration | > process          | 'healthcheck'
18:17:39 | 1      | configuration | . run              | 'python3' '-m' 'exabgp' 'healthcheck' '--no-syslog' ...
18:17:39 | 1      | configuration | . encoder          | 'text'
18:17:39 | 1      | configuration | < process          | 
18:17:39 | 1      | configuration | > template         | 
...
18:17:39 | 1      | reactor       | peer definition identical, updating peer routes if required for neighbor xxx ... family-allowed in-open
18:17:39 | 1      | reactor       | loaded new configuration successfully
18:17:39 | 1      | process       | terminating process healthcheck
INFO[healthcheck] send announces for EXIT state to ExaBGP
18:17:39 | 1      | process       | forked process healthcheck
INFO[healthcheck] send announces for UP state to ExaBGP
18:17:41 | 1      | process       | command from process healthcheck : announce route xxx 
```

A ps then shows:

```
nobody    741593  0.0  0.0  12128 11200 pts/0    S    14:24   0:00 /usr/local/bin/python3 -m exabgp healthcheck ...
nobody    741647  0.0  0.0  12128 11268 pts/0    S    14:24   0:00 /usr/local/bin/python3 -m exabgp healthcheck...
nobody    741872  0.0  0.0  12128 11284 pts/0    S    14:25   0:00 /usr/local/bin/python3 -m exabgp healthcheck ...
nobody    741915  0.0  0.0  12488 11568 pts/0    S    14:25   0:00 /usr/local/bin/python3 -m exabgp healthcheck ...
... many more
```

**Expected behavior**

[...]

PR #1029, opened 2021-03-15
When reloading, force kill old process after a timeout

vincentbernat:
This is related to #1006. If the process does not shutdown with
SIGTERM, use SIGKILL.

Note that the `timeout` argument for `wait()` is an addition for
Python 3. If we still need to support Python 2, the easiest way would
be to just not to do that and instead use `kill()` instead of
`terminate()` to not have this issue.

Also revert the change in `healthcheck.py` checking for the PPID.
AFAIK, the process is not reparented.

### DIFF
--- a/src/exabgp/application/healthcheck.py
+++ b/src/exabgp/application/healthcheck.py
@@ -496,7 +496,7 @@ def loop(options):
 
     signal.signal(signal.SIGTERM, sigterm_handler)
 
-    while os.getppid() != 1:
+    while True:
         checks, state = one(checks, state)
 
         try:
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Thank you vincent, the patch is much appreciated. 

The wait will cause the async engine to pause for possibly 2 seconds, if many processes are running and the problem is global, then a "long'ish" pause could happen. We could perhaps perform the killing using fork(). Do you agree?

I am happy to take this code and patch it a bit later.

vincentbernat:
❦ 15 mars 2021 14:21 -07, Thomas Mangin:

> The wait will cause the async engine to pause for possibly 2 seconds,
> if many processes are running and the problem is global, then a
> "long'ish" pause could happen. We could perhaps perform the killing
> using fork(). Do you agree?

It's already in a thread. But I suppose you noticed that since you've
merged nonetheless?

PR #1068, opened 2022-02-24
Add Support for Prefix-SID type-5 and type-6 

proelbtn:
This patch includes:

- remove deprecated Prefix-SID definition (type-2 and type-4)
- implement Prefix-SID type-5 and type-6
- add new configuration for Prefix-SID type-5 and type-6

reference: https://datatracker.ietf.org/doc/draft-ietf-bess-srv6-services/11/


thomas-mangin:
Thank you for this work, much appreciated.

In order to be able to merge this work, a JSON representation of the objects need to be added to the classes: all the JSON functions were left as `TODO`.

proelbtn:
Yes, I'm considering JSON format for them and I was so tired last night that I fell asleep.

proelbtn:
@thomas-mangin I added JSON representation and fixed some small things. It's ok to be reviewed it.

thomas-mangin:
Thank you I will review. There was no rush - I hope you are rested now 🥱

thomas-mangin:
Ok, it seems that it is breaking backwards compatibility with the previous type names (VPN and l3vpn) so it should be included in `doc/CHANGELOG.rst` and the RFC supported on `https://github.com/Exa-Networks/exabgp/wiki/RFC-Information` also but otherwise LGTM.

If you want to add the CHANGELOG entry to the patch set, I will then merge 👍

proelbtn:
@thomas-mangin updated CHANGELOG

PR #1098, opened 2022-06-14
healthcheck.py - BGP Add Path

gbe0:
Hi,

I would like to request BGP add path support for `healthcheck.py`. The server running ExaBGP may be separate from the devices being health checked; when this is the case it can be useful to have BGP add path support to allow equal cost multipath if there are multiple routes for the same prefix.


thomas-mangin:
LGTM - thank you for the patch!

PR #1099, opened 2022-06-14
Add neighbor filtering to healthcheck.py

gbe0:
In healthcheck.py, it would be useful to specify an option for a set of neighbors to announce the route to. Currently route advertisements are sent to ExaBGP like this:

```
announce route 192.0.2.1/32 next-hop 192.0.2.250 ...
```

As an option, it would be nice to be able to send announcements like this:

```
neighbor 192.0.2.200 announce route 192.0.2.1/32 next-hop 192.0.2.250 ...
neighbor 192.0.2.200, neighbor 192.0.2.201 announce route 192.0.2.2/32 next-hop 192.0.2.251 ...
```

This patch adds the `--neighbor` option to do exactly this.


thomas-mangin:
LGTM (again), thank you (again) 😉

PR #1127, opened 2022-11-28
Refactor how config originated routes are deleted from outgoing RIB t…

longmalx:
Previous the old routes from the configuration were deleted from the peer reactor, leading to the RIB become out of date if the neighbor was offline when an update or sequence of updates was made.

Refactored to this so that any deleted configuration routes are removed from RIB during configuration reload rather than from the reactor.

Tested this with a variety of normal and FlowSpec routes and the RIB is correct all configuration reloads.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,8 @@ Version explained:
  - bug   : increase on bug or incremental changes
 
 Version 4.2.22
+ * Fix: route reload for offline neighbors #1126
+   patch: Malcolm Dodds
 
 Version 4.2.21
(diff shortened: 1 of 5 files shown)


longmalx:
I've noticed that just updating the next hop or community doesn't seem to be handled correctly - will investigate.

thomas-mangin:
Thank you very much, Malcolm. Are all the tests still passing? in particular, the encoding tests in `./qa/bin/functional`? If so, I will merge.

longmalx:
I have a fixed up patch which is working correctly with my manual testing - I will try the functional tests.

Issue #1126, opened 2022-11-15
'adj-rib out' is incorrect after reloading configuration with an offline router


longmalx:
**Bug Description**

When a neighbor is unavailable during configuration changes, we have noticed that the adj-rib may contain routes which have been removed from the configuration. These remain even when the neighbor becomes available again.

All route changes are performed by updating exabgp.conf and issuing a reload.

**To Reproduce**

I have 2 configuration files (contained in [exabgp.conf.zip](https://github.com/Exa-Networks/exabgp/files/10012105/exabgp.conf.zip)):
- exabgp.conf.empty (2 neighbors (1 available, 1 not) and no routes
- exabgp.conf.routes (2 neighbors (1 available, 1 not) and no routes

I run commands to set the configuration to empty, restart, add the routes, reload, remove routes, reload
```
[root@edx-cms exabgp]# cp exabgp.conf.empty /etc/exabgp/exabgp.conf[root@edx-cms exabgp]# systemctl restart exabgp
[root@edx-cms exabgp]# cp exabgp.conf.routes /etc/exabgp/exabgp.conf
[root@edx-cms exabgp]# systemctl reload exabgp
[root@edx-cms exabgp]# exabgpcli show adj-rib out extensive;
neighbor 192.168.200.254 local-ip 0.0.0.0 local-as 65000 peer-as 65000 router-id 10.10.243.170 family-allowed in-open ipv4 flow flow
destination-ipv4 172.16.1.1/32 extended-community redirect:666:666
neighbor 192.168.200.253 local-ip 0.0.0.0 local-as 65000 peer-as 65000 router-id 10.10.243.170 family-allowed in-open ipv4 flow flow
destination-ipv4 172.16.1.1/32 extended-community redirect:666:666
[root@edx-cms exabgp]# cp exabgp.conf.empty /etc/exabgp/exabgp.conf
[root@edx-cms exabgp]# systemctl reload exabgp
[root@edx-cms exabgp]# exabgpcli show adj-rib out extensive;
neighbor 192.168.200.253 local-ip 0.0.0.0 local-as 65000 peer-as 65000 router-id 10.10.243.170 family-allowed in-open ipv4 flow flow
destination-ipv4 172.16.1.1/32 extended-community redirect:666:666
[root@edx-cms exabgp]# exabgpcli show neighbor summary
Peer            AS        up/down state       |     #sent     #recvd
192.168.200.254 65000     0:01:33 established           6          4
```
[...]

PR #1128, opened 2022-11-28
Refactor how config routes are deleted from outgoing RIB to handle offline routers

longmalx:
This resolves issue  #1126.

Previously the old routes from the configuration were deleted from the peer reactor, leading to the RIB become out of date if the neighbor was offline when an update or sequence of updates was made.

Refactored to this so that any deleted configuration routes are removed from RIB during configuration reload rather than from the reactor.

Tested this with a variety of normal and FlowSpec routes and the RIB is correct all configuration reloads.
I have run all the QA functional tests successfully.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,8 @@ Version explained:
  - bug   : increase on bug or incremental changes
 
 Version 4.2.22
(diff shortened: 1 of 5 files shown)


thomas-mangin:
I have merged the patch but now wonder if there will be an issue with API routes? I need to check.

PR #1133, opened 2022-12-20
Minor fix to make docker build work

paddy01:
docker build failed due to error in Dockerfile.remote - fixed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,10 +28,8 @@ RUN chown exa /run/exabgp.out
 RUN chmod 600 /run/exabgp.in
 RUN chmod 600 /run/exabgp.out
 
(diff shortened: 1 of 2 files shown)


thomas-mangin:
COPY was not supported by Podman last time I checked, hence the RUN echo syntax.

thomas-mangin:
Thank you @paddy01 - if it fixes your problem, please feel free to close the issue/PR.

paddy01:
Have to reopen this as it doesn't work in a raw docker build .. it does work under docker compose build tho which is odd. -- I'll remake it to use echo for both instead to make it compliant, sorry!

paddy01:
This should do it.

PR #1142, opened 2023-03-13
Support for BGP-MUP SAFI and Extended Community

takehaya:
This PR enables exaBGP to support BGP-MUP SAFI and Extended Community defined in [draft-mpmz-bess-mup-safi-02](https://datatracker.ietf.org/doc/html/draft-mpmz-bess-mup-safi-02).

SAFI value 85 is assigned by IANA as "BGP-MUP SAFI" https://www.iana.org/assignments/safi-namespace/safi-namespace.xhtml

BGP Transitive Extended Community Types 0x0c is assigned by IANA as "SRv6 MUP Extended Community" https://www.iana.org/assignments/bgp-extended-communities/bgp-extended-communities.xhtml

Usage: https://gist.github.com/takehaya/8170122cb7ff4c710775d02123903e70


thomas-mangin:
Hello @takehaya - Thank you very much for this patch, it looks pretty good! I will give it a longer review a bit later on.

At first glance, and from previous experience, the main thing I can not change once the code is accepted is the name of the JSON attributes. Right now, I see a mix of spaces, underscores (but no hyphen 👍). I think the user experience would be better with consistent naming and usage of prefixes and suffixes.

Also, I note the use of "parsed: true". I am not sure this is providing much. Was there a good reason for you to add this when checking for anything more than "raw" may achieve the same result?

Once again thank you for this work.

takehaya:
Hi @thomas-mangin -san
Thank you for your quick review! 👍 
I confirmed that the json output was strange, so I fixed it. [fix json format](https://github.com/Exa-Networks/exabgp/pull/1142/commits/93ea159f1b397b3ace6049833164df36d2283fb6)
plz check code 🙏

thomas-mangin:
Hi @takehaya-san, arigato gozaimashita.

takehaya:
Oh!! nice japanese! 😲
I am very happy that you spoke to me in Japanese👏 Arigato gozaimasu(ありがとうございます！)

I got it. Please few a wait that i write by test code...:)

thomas-mangin:
I was told how to correctly say thank you by an old lady in a Tokyo shop 8 years ago when I visited the country for 10 days. I spend a week in Tokyo and then visited Kyoto, Osaka, the Nara Park and Himeji Castle - everyone was so nice, one of my best holidays :-)

takehaya:
that "Arigato" is perfect 👏 so nice experience...!
Wow. I live in Kyoto :) The cherry blossoms should be beautiful in Kyoto soon🌸 

Actually, I'm planning to bring this code to IETF Hackathon 116 in Japan Yokohama this month.
That's why I was doing this.

https://wiki.ietf.org/en/meeting/116/hackathon
> BGP-MUP SAFI Implementation and Interop

---
I tried adding a test. please confirm  🙏 
(and add the case of v6 config)

thomas-mangin:
Yes, I came at the end of March, and the trees were beautiful.

The code is merged 🥳 🍾

Thank you for all this work. I was wondering how you found your experience working with the code. Any issues?

thomas-mangin:
One last thing. I see that in the log, the routes are presented as "insert mup:isd::100:100:10.0.1.0/24" - do you mind if it is changed to be the same syntax as the annonouce ipv4 section ?

takehaya:
Please come to Japan again if you have the chance :)

Thank you for your quick and polite review! 🍷 

It was pretty well managed and readable code. It was great! (I liked that the documentation was fairly well maintained!)
by the way, there were three main problems in doing this work.
1. the fmt rules for CI lint and make were different. when we applied fmt, we were surprised to see a large number of changes when we applied fmt.
2. I was worried because I didn't know what kind of syntax to use when adding a new function.
3. I'm having trouble finding documentation about test vectors.

> I see that in the log, the routes are presented as "insert mup:isd:💯100:10.0.1.0/24" - do you mind if it is changed to be the same syntax as the annonouce ipv4 section ?

of course !

PR #1180, opened 2023-09-27
feat(neighbor): support announce all processes

ahmet2mir:
Hi,

As discused on slack, currently there is no way to tell neighbours to announce all defined processes.

Adding "all" inside api.processes allow announcing all of them.


thomas-mangin:
Good idea and the only issue with with patch would be the missing second run file. That said, I would rather have a new keyword “processes-matching” taking a regex expression which would be used to decide which process names to include and add a alias for processes called processes-named perhaps too

PR #1185, opened 2023-10-17
Update extended-nexthop.conf to use documentation prefixes.

axu-trex:


thomas-mangin:
Thank you for this patch, but I will refuse it as it is now: the IPs are as it is to allow local testing.

The ASN and IP announced can be changed. 

If the patch is modified to only change the ASN and IP, it can be then merged.

thomas-mangin:
Thank you @axu-trex 

Sorry, I only noticed now that if you change this file, then https://github.com/Exa-Networks/exabgp/blob/main/qa/encoding/conf-extended-attributes.msg also need to be changed to reflect the new routes sent as otherwise, the CI will fail.

Please look at the `./qa/bin/functional` program in the repository which runs our tests.
with `./qa/bin/functional encoding --help`.

This program with "encoding" will start a client using one of the configuration files in the ".ci" file and a server to validate the output of the BGP packets generated by exabgp.

You can find how to run the "client" and "server" for any test, and with `--list` you can get the "id" of the test to run.

thomas-mangin:
I also acknowledge that the phrasing in my initial response may not have been overly friendly, and I am appreciative of your engagement.

thomas-mangin:
Thank you for pointing the possible use of rfc ip ranges in files. I will take it in consideration going forward.

PR #1189, opened 2023-11-24
Fixes flowspec redirect

louis-6wind:
See commit logs


thomas-mangin:
Thank you very much @louis-6wind !

PR #1190, opened 2023-12-06
Fix ipv6 redirect 4.2

louis-6wind:
Backport of https://github.com/Exa-Networks/exabgp/pull/1189


louis-6wind:
@thomas-mangin Can we have a new tag with this patch ? 4.2.22 ?

thomas-mangin:
I would need to first remove a patch I accepted and reapply it after the update but is is possible in the next few days

PR #1193, opened 2023-12-16
mup: add Source Address to Type 1 ST Route 

takehaya:
This patch adds Source Address to Type 1 ST Route as defined in draft-mpmz-bess-mup-safi-03.
See https://datatracker.ietf.org/doc/html/draft-mpmz-bess-mup-safi-03#section-3.1.3


thomas-mangin:
Very nice. Could I please ask for a separate test rather than extending the current one.

src/exabgp/bgp/message/update/nlri/mup/t1st.py
```diff
@@ -165,8 +185,24 @@ def unpack(cls, data, afi):
...
         endpoint_ip_len = data[size]
         size += 1
-        endpoint_ip = IP.unpack(data[size:])
-        return cls(rd, ipprefix_len, ipprefix, teid, qfi, endpoint_ip_len, endpoint_ip, afi)
+
+        if endpoint_ip_len in [32, 128]:
+            endpoint_ip = IP.unpack(data[size : size + endpoint_ip_len])
+            size += endpoint_ip_len
+        else:
+            raise RuntimeError('endpoint ip length is not 4 or 16')
```
thomas-mangin, line 193: can the runtime error be a bit more specific about which endpoint ip is wrong to help people when they look at the logs.

src/exabgp/bgp/message/update/nlri/mup/t1st.py
```diff
@@ -165,8 +185,24 @@ def unpack(cls, data, afi):
...
+
+        source_ip_size = datasize - size
+        if 0 < source_ip_size:
+            source_ip_len = data[size]
+            size += 1
+            if source_ip_len in [4, 16]:
+                source_ip = IP.unpack(data[size : size + source_ip_len])
+                size += source_ip_len
+            else:
+                raise RuntimeError('source ip length is not 4 or 16')
```
thomas-mangin, line 203: Same here about the error message, a bit more specific please.

thomas-mangin:
Thank you for your work, if you can please update the PR with the suggested changes, I will then happily merge.

takehaya:
@thomas-mangin san Thank you for your kind review!

> Very nice. Could I please ask for a separate test rather than extending the current one.

What tests does this expect to add?I would like to know if you would like 🙏

thomas-mangin:
You are welcome.

The test you wrote is good but it is using an existing file. If you create a new configuration and test files instead, it keeps the new test separate.

The configuration file needs to be created in etc/exabgp and a .ci and .msg file created in qa/encoding. qa/bin/functional can be used to run the new test.

takehaya:
thanks comment :)
i try fixed that because your kind feedback. pls continue review 🙏

src/exabgp/bgp/message/update/nlri/mup/t1st.py
```diff
@@ -77,23 +85,30 @@ def __eq__(self, other):
...
+            and self.source_ip_len == other.source_ip_len
+            and self.source_ip == other.source_ip
         )
 
     def __ne__(self, other):
         return not self.__eq__(other)
 
     def __str__(self):
-        return "%s:%s:%s%s:%s:%s:%s:%s" % (
+        s = "%s:%s:%s%s:%s:%s:%s%s" % (
```
thomas-mangin, line 96: Please can you have two code path: one when source_ip_len and source_ip is not set which does not change how the object is represented, and a other branch where you can add your information, keeping the current fields unchanged and adding the two new extra ones.

What you are proposing here seems to once more change how the data would be serialised which may break code from current users.

takehaya: thank review!
I don't know if I understand the meaning well.
Is this what you're saying?
[14dd27a](https://github.com/Exa-Networks/exabgp/pull/1193/commits/14dd27a8cdce95f3626336d8408662e778782d1c)

thomas-mangin: The output of the function should be unchanged when parsing packets which do not match the new code.

thomas-mangin: fixed and merged

takehaya:
@thomas-mangin -san thanks! 🙏

PR #1196, opened 2024-01-15
implement --debounce flag in healthcheck.py

Mateusz-Laczny:
Hello,

I've implemented a new flag in `healthcheck.py`, called `debounce`, that allows to reduce the number of sent announcements - when it's set, the announcements are sent only when a state change is detected. The flag is disabled by default.


thomas-mangin:
Thank you for your contribution.

PR #1200, opened 2024-03-12
healthcheck: fix toconsole condition logging

clwluvw:
Log to console either hasattr() or isatty()

### DIFF
--- a/src/exabgp/application/healthcheck.py
+++ b/src/exabgp/application/healthcheck.py
@@ -174,8 +174,7 @@ def setup_logging(debug, silent, name, syslog_facility, syslog):
         sh.setFormatter(logging.Formatter("{0}[{1}]: %(message)s".format(healthcheck_name, os.getpid())))
         logger.addHandler(sh)
     # To console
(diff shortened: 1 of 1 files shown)


vincentbernat:
This doesn't look correct. The check requires the `isatty` attribute to exist to call it. If I remember correctly, it may not exist on Windows, hence the check. With your change, if the attribute does not exist, you call it, which will trigger an exception. And if it exists, you don't really check this is a tty.

clwluvw:
@vincentbernat This is now not working inside a docker container. Do you have any alternative way to solve it?

vincentbernat:
Well, I suppose nowaday, when running daemonized, stderr is set to `/dev/null`, so it shouldn't be a problem to write to it even when this is a not tty, so we may just replace `toconsole` by `not silent`. People can still opt out that with `--silent`.

clwluvw:
Yeah, that was my first approach as well - but thought maybe better to keep it both. I changed it.

vincentbernat:
Just do `if not silent:` instead.

clwluvw:
Done. PTAL.

vincentbernat (approved):

PR #1206, opened 2024-04-14
source-interface parameter for peering

mxyns:
running exabgp in a vrf context is buggy on some OSes
this PR adds an optional `source-interface` parameter to the peer configuration
this allows selecting over which interface the the session with the peer is established
the interface may be in a vrf


thomas-mangin:
Thank you very much for this PR and the time you spend adding this feature, LGTM.

Issue #1199, opened 2024-03-12
healthcheck: Loopback IP Parsing


clwluvw:
** Bug Report **

**Describe the bug**

When having two ips with labels like `labelname` and `labelname-something`, specifying `--label labelname` would pick both `labelname` and `labelname-something`.

**Expected behavior**

Only the desired label be chosen.

**Additional context**

It must be probably because of the `startswith` func here (https://github.com/Exa-Networks/exabgp/blob/main/src/exabgp/application/healthcheck.py#L221)

clwluvw:
I see a backslash in the end of the label from `/sbin/ip -o address show dev lo` output.

vincentbernat:
I don't remember if it was the reason I have used `startswith`. Maybe the right fix would be:
```
if lmo.groupdict().get("label", "").rstrip("\\") == label:

vincentbernat:
Sorry, looking at the documentation string, this is something expected. Labels need to be unique, so that's why there is a `startswith()`. If you want to use several loopbacks, you label them `haproxy1`, `haproxy2`, `haproxy3` and you use `--label haproxy`.

clwluvw:
hmm - maybe a better approach was to support regex as an arg so this wouldn't be confusing. I mean now you can't use it unless you change the names completely.

thomas-mangin:
@clwluvw if you wish to add an option ( --use-interface-numbering ?? ) to have the code use numbering while keeping backward compatibility for previous users when not used, I will happily have a look at it and merge it if it looks good.

As I develop on mac, testing Linux networking is a pain (and yes, I know about VM and containers). As I agree with @vincentbernat that this is not a big problem, I do not feel like working on it.

Still, if you care enough to put the work into it, I will perform the code review on behalf of Vincent. If you do not wish to do so, please fell free to close the issue.

PR #1215, opened 2024-06-23
healthcheck: add support interface numbering

clwluvw:
So far loopback ips was looking for the corresponding interface by checking whether the label starts with the desired input. By providing --use-interface-numbering it checks whether the label exactly matches the desired input.

Fixes: #1199

### DIFF
--- a/src/exabgp/application/healthcheck.py
+++ b/src/exabgp/application/healthcheck.py
@@ -96,6 +96,7 @@ def setargs(parser):
     g.add_argument("--no-ip-setup", action="store_false", dest="ip_setup", help="don't setup missing IP addresses")
     g.add_argument("--dynamic-ip-setup", default=False, action="store_true", dest="ip_dynamic", help="delete existing loopback ips on state down and " "disabled, then restore loopback when up")
     g.add_argument("--label", default=None, help="use the provided label to match loopback addresses")
+    g.add_argument("--label-exact-match", default=False, action="store_true", help="use the provided label to exactly match loopback addresses, not a prefix match")
     g.add_argument("--start-ip", metavar='N', type=int, default=0, help="index of the first IP in the list of IP addresses")
     g.add_argument("--up-metric", metavar='M', type=int, default=100, help="first IP get the metric M when the service is up")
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Please let me know when you are happy that the code is fine and I will review for merge.

clwluvw:
@thomas-mangin I guess that should be it. Happy to take your review.

src/exabgp/application/healthcheck.py
```diff
@@ -180,15 +181,15 @@ def syslog_address():
...
+def loopback_ips(label, label_only, interface_numbering):
     """Retrieve loopback IP addresses"""
     logger.debug("Retrieve loopback IP addresses")
     addresses = []
 
     if sys.platform.startswith("linux"):
         # Use "ip" (ifconfig is not able to see all addresses)
         ipre = re.compile(r"^(?P<index>\d+):\s+(?P<name>\S+)\s+inet6?\s+" r"(?P<ip>[\da-f.:]+)/(?P<mask>\d+)\s+.*")
-        labelre = re.compile(r".*\s+lo:(?P<label>\S+).*")
+        labelre = re.compile(r".*lo:(?P<label>[^\\]+).*")
```
vincentbernat, line 192: This would match "hello" interface as well. Why change the regex?

vincentbernat:
The name of the option is odd, isn't it? This could be `--label-exact-match` instead.

src/exabgp/application/healthcheck.py
clwluvw: I wanted to drop the last `\` from the label. this is an example output in Linux:
```
1: lo    inet 127.0.0.1/8 scope host lo:label\       valid_lft forever preferred_lft forever
```
which the previous regex would still carry the `\`. I don't know what do you mean about the "hello" interface. Can you give me an example or suggest a regex for this case?

vincentbernat: Removing `\s+` opens the possibility to match more than `lo:something`. Use `[^\\\s]` to also include blanks, just in case the backslash is not always there.

thomas-mangin:
@vincentbernat LGTM

thomas-mangin:
@vincentbernat thank you for your time and the review !

PR #1220, opened 2024-07-09
Add missing `exc` exception variable for qa/bin/functional

ton31337:
File "./exabgp/./qa/bin/functional", line 155, in collect
    self.message = str(exc)
UnboundLocalError: local variable 'exc' referenced before assignment


thomas-mangin:
Thank you.

PR #1221, opened 2024-07-09
Do not send hostname/domainname (fqdn capability) by default

ton31337:


ton31337:
@thomas-mangin Thanks for merging previous commits, I added here a bit more checks: like if the domain is unset, but only hostname is set, it should to continue sending only hostname part.

thomas-mangin:
Sure. LGTM.

ton31337:
Thanks!

thomas-mangin:
Can you please add a test in the test for this draft please?

ton31337:
I think yes, let me figure out how to do it.

thomas-mangin:
`./qa/bin/functional encoding --edit 1` will show you how test one is setup. You can get inspiration from there

thomas-mangin:
`exabgp -d` will print what it sent, and the bgp software in `qa/sbin` does as well, I normally use `./qa/bin/function encoding --server 1` with a `--client 1` and then can copy the output from the server (bgp) in the file.

Normally we do only care about UPDATE, so you may have to look at the code on how to deal with OPEN .. or add the feature, not sure.

Some msg files have a comment as the first line which is a control from the bgp deamon on how to behave.

PR #1222, opened 2024-07-09
Test if capability message is received as expected

ton31337:


thomas-mangin:
can I please ask you to reformat the function in another PR so I can see what changed please?

thomas-mangin:
As the bgp program is changed bgp-3.6 will also be updated to match the changes otherwise I will have no chance to make Python 3.6 build stable - ATM they fail once in a while but I do not know why (and I need to build a old Ubuntu to debug) so I have left it for later.

thomas-mangin:
Forget my comment about formatting ..

ton31337:
I can drop that reformatting commit for sure, should I drop it or leave it as it is now?

ton31337:
Will do in a minute ;)

thomas-mangin:
ok, it does not allow to test for an OPEN and then an UPDATE - that's ok, it can be added later. It does not break anything.

ton31337:
Gh, don't have 3.6 right now...

PR #1233, opened 2024-09-04
Update exabgp master to 4.2.22

k-304:


thomas-mangin:
No worries. main/master is ahead of 4.2.22 as many features have not been backported. Mostly as some changes are not fully backward compatible. It is at a point where it would justify a 5.0.0 release but I have been holding it back.

thomas-mangin:
For clarity, there is no fix in 4.2.22 which is not already in main. Some backports may have required a different implementation because the code base diverged over time. main can be used in production as it sees the same care for stability as 4.2.

PR #1234, opened 2024-09-05
Implement RFC6514 MCAST-VPN (incomplete)

jcpvdm:
https://datatracker.ietf.org/doc/html/rfc6514
This commit starts support for MCAST-VPN NLRI. Route Types added:
+ 5 - Source Active A-D route
+ 6 - Shared Tree Join route
+ 7 - Source Tree Join route

This came out of my need to perform tests for some MCAST-VPN route types. exabgp fit nicely for the task :)
Note there was some reformatting done on unrelated code by black.


src/exabgp/bgp/message/update/nlri/mvpn/sharedjoin.py
```diff
@@ -0,0 +1,114 @@
...
+# +-----------------------------------+
+# |  Multicast Group   (variable)     |
+# +-----------------------------------+
+
+
+@MVPN.register
+class SharedJoin(MVPN):
+    CODE = 6
+    NAME = "C-Multicast Shared Tree Join route"
+    SHORT_NAME = "Shared-Join"
```
thomas-mangin, line 29: This name may be used when the route is printed and may need to be lower case?

src/exabgp/bgp/message/update/nlri/mvpn/sharedjoin.py
```diff
@@ -0,0 +1,114 @@
...
+            and self.CODE == other.CODE
+            and self.rd == other.rd
+            and self.source == other.source
+            and self.group == other.group
+        )
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __str__(self):
```
thomas-mangin, line 51: Does this represent the route in a way which could be parsed by the configuration code?

thomas-mangin:
Hello @jcpvdm - at first glance, it is an excellent commit. Thank you very much. I have two observation question above, which I would appreciate if you could consider.

src/exabgp/bgp/message/update/nlri/mvpn/sharedjoin.py
jcpvdm: You mean NAME, SHORT_NAME or both? I tried to follow example from other nlri code (like mup and evpn), but didn't notice a pattern regarding lower-case use.

Notice SHORT_NAME is already lower-cased here when printing by _prefix():
https://github.com/Exa-Networks/exabgp/blob/279a289ba4cd5a5e6ab2a6c98d494b9d6fb33e9c/src/exabgp/bgp/message/update/nlri/mvpn/sharedjoin.py#L51-L52
https://github.com/Exa-Networks/exabgp/blob/279a289ba4cd5a5e6ab2a6c98d494b9d6fb33e9c/src/exabgp/bgp/message/update/nlri/mvpn/nlri.py#L60-L61

This is the print output you mean right?
```bash
reactor       connected to peer-1 with outgoing-1 10.99.12.6-10.99.12.5
processes     ipv4 added to neighbor 10.99.12.5 local-ip 10.99.12.6 local-as 65000 peer-as 65000 router-id 32.32.32.32 family-allowed in-open : mvpn:shared-join::65000:99999:65000:10.99.199.1:239.251.255.228 extended-community target:192.168.94.12:5
processes     ipv4 added to neighbor 10.99.12.5 local-ip 10.99.12.6 local-as 65000 peer-as 65000 router-id 32.32.32.32 family-allowed in-open : mvpn:source-join::65000:99999:65000:10.99.12.2:239.251.255.228 extended-community target:192.168.94.12:5
processes     ipv6 added to neighbor 10.99.12.5 local-ip 10.99.12.6 local-as 65000 peer-as 65000 router-id 32.32.32.32 family-allowed in-open : mvpn:shared-join::65000:99999:65000:fd00::1:ff0e::1 extended-community target:192.168.94.12:5
processes     ipv6 added to neighbor 10.99.12.5 local-ip 10.99.12.6 local-as 65000 peer-as 65000 router-id 32.32.32.32 family-allowed in-open : mvpn:source-join::65000:99999:65000:fd12::2:ff0e::1 extended-community target:192.168.94.12:5
processes     ipv6 added to neighbor 10.99.12.5 local-ip 10.99.12.6 local-as 65000 peer-as 65000 router-id 32.32.32.32 family-allowed in-open : mvpn:sourcead::65000:99999:fd12::4:ff0e::1 extended-community target:65000:99999
processes     ipv4 added to neighbor 10.99.12.5 local-ip 10.99.12.6 local-as 65000 peer-as 65000 router-id 32.32.32.32 family-allowed in-open : mvpn:sourcead::65000:99999:10.99.12.4:239.251.255.228 extended-community target:65000:99999
```

ton31337 (commented):
From the readability perspective, I'd rather use a normal (full) name `multicast-vpn`, instead of `mcast-vpn`.

src/exabgp/bgp/message/update/nlri/mvpn/sharedjoin.py
jcpvdm: I didn't get this one. can you share an example?

jcpvdm:
> From the readability perspective, I'd rather use a normal (full) name `multicast-vpn`, instead of `mcast-vpn`.

You mean replace all instances of "mcast-vpn" by "multicast-vpn" in the code?

ton31337:
> > From the readability perspective, I'd rather use a normal (full) name `multicast-vpn`, instead of `mcast-vpn`.
> 
> You mean replace all instances of "mcast-vpn" by "multicast-vpn" in the code?

At least in configuration mode :) (seeing as an operator).

thomas-mangin:
I will need to write some tests and may have to perform a non-backwards compatible change to have what I just asked. I will think about it as I have not enforced this consistently.

Merging. Happy to accept follow up fixes.

thomas-mangin:
Thank you very much for this work. It is much appreciated. Thank you, @ton31337, too, for the feedback.

Issue #909, opened 2019-08-11
JSON output for API command responses


Vascko:
##### Summary
Hey Guys,
I'm not quite sure at this point if this is a bug or feature request. Please clarify: I configured the JSON format in every place I could find but ExaBGP responses to commands like "show adj-rib out" seem to always be in text format. No matter if the command has been sent through the native CLI tool or on the STDOUT of a process. Is this intended? If yes, implementing text format parsing in my helper processes would be a lot more complicated than if the response could be configured to be JSON. The JSON output on other API events like updates or neighbor changes works fine but I'm interested in the actual API responses to commands coming from a helper process.

##### OS

```
CentOS 7.6
```

```
Linux localhost 4.14.35-1844.4.5.el7.x86_64 #2 SMP Tue Apr 9 00:29:47 PDT 2019 x86_64
```

##### Version

```
exabgp --version
ExaBGP : 4.1.2-d006a34a
Python : 3.6.2 (default, Jul 20 2017, 14:22:24)  [GCC 4.4.7 20120313 (Red Hat 4.4.7-18)]
Uname  : Linux localhost 4.14.35-1844.4.5.el7.x86_64 #2 SMP Tue Apr 9 00:29:47 PDT 2019 x86_64
Root   : /usr/local/python
```

##### Installation

```
setup.py install
```

##### Environment 

```
Output empty (didn't change the environment from defaults)
```

##### Configuration

```
process bgplus {
  run exa_api;
  encoder json;
}

neighbor 192.168.20.5 {
  description "gateway";
  router-id 192.168.20.100;
  local-address 192.168.20.100;
  local-as 12345;
  peer-as 12346;
  hold-time 3;
  manual-eor false;

  passive false;
  group-updates true;
  auto-flush true;
  adj-rib-in true;
  adj-rib-out true;
  md5-base64 auto;
  md5-ip "192.168.20.100";

  capability {
    asn4 enable;
    route-refresh disable;
    graceful-restart disable;
    nexthop disable;
    add-path disable;
    multi-session disable;
    operational disable;
    aigp disable;
  }
  family {
    ipv4 unicast;
  }
  nexthop {
  }
  add-path {
  }
  api ExaAPI {
    processes [ exa_api ];
  }
}
```

##### Program output

This is the logging output from my helper program that prints the data it gets back on its STDIN from ExaBGP
```
ExaAPI - INFO: Withdrawing announcement for 1.1.1.1/32
[...]

PR #1240, opened 2024-10-16
Jsonify show neighbor,adj-rib-in,adj-rib-out cmds output

vish-321:
This PR updates ExaBGP to optionally provide JSON output for API commands such as show neighbor, adj-rib-in, and adj-rib-out.

Existing Commands:
```
show neighbor [optional neighbor ip] summary
show neighbor [optional neighbor ip] extensive
show neighbor [optional neighbor ip] configuration
show adj-rib in [extensive] (to be integrated with neighbor)
show adj-rib out [extensive] (to be integrated with neighbor)
```

Example Usage for JSON Output:
```
show neighbor [optional json] [optional neighbor ip] summary
show neighbor [optional json] [optional neighbor ip] extensive
show adj-rib in [optional json] [extensive] (to be integrated with neighbor)
show adj-rib out [optional json] [extensive] (to be integrated with neighbor)
```
Examples:
```
show neighbor json summary
show neighbor json extensive
show adj-rib out json
```

This should also resolve the issue detailed in https://github.com/Exa-Networks/exabgp/issues/909

### DIFF
--- a/src/exabgp/bgp/neighbor.py
+++ b/src/exabgp/bgp/neighbor.py
@@ -173,19 +173,31 @@ class Neighbor(dict):
         self.messages = deque()
         self.refresh = deque()
 
-    def name(self):
+    def name(self, json=False):
         if self['capability']['multi-session']:
             session = '/'.join("%s-%s" % (afi.name(), safi.name()) for (afi, safi) in self.families())
         else:
             session = 'in-open'
-        return "neighbor %s local-ip %s local-as %s peer-as %s router-id %s family-allowed %s" % (
-            self['peer-address'],
-            self['local-address'] if self['peer-address'] is not None else 'auto',
-            self['local-as'] if self['local-as'] is not None else 'auto',
-            self['peer-as'] if self['peer-as'] is not None else 'auto',
-            self['router-id'],
-            session,
-        )
+        data = {
+            "neighbor": self.peer_address,
+            "local-ip": self.local_address if self.peer_address is not None else 'auto',
+            "local-as": self.local_as if self.local_as is not None else 'auto',
+            "peer-as": self.peer_as if self.peer_as is not None else 'auto',
+            "router-id": self.router_id,
+            "family-allowed": session
+        }
+
+        if json:
+            return data
+        else:
+            return "neighbor %s local-ip %s local-as %s peer-as %s router-id %s family-allowed %s" % (
+                data["neighbor"],
+                data["local-ip"],
+                data["local-as"],
+                data["peer-as"],
+                data["router-id"],
+                data["family-allowed"]
+            )
 
     def families(self):
         # this list() is important .. as we use the function to modify self._families
--- a/src/exabgp/reactor/api/command/neighbor.py
+++ b/src/exabgp/reactor/api/command/neighbor.py
@@ -32,38 +32,28 @@ def _pr(value):
     return '%s' % value
 
 
-def _addpath(send, receive):
-    if send and receive:
-        return "send/receive"
-    if send:
-        return "send"
-    if receive:
-        return "receive"
-    return "disabled"
-
-
 class Neighbor(object):
     extensive_kv = '   %-20s %15s %15s %15s'
     extensive_template = """\
 Neighbor %(peer-address)s
 
-    Session                         Local
+	Session                         Local
 %(local-address)s
 %(state)s
 %(duration)s
 
-    Setup                           Local          Remote
+	Setup                           Local          Remote
 %(as)s
 %(id)s
 %(hold)s
 
-    Capability                      Local          Remote
+	Capability                      Local          Remote
 %(capabilities)s
 
-    Families                        Local          Remote        Add-Path
+	Families                        Local          Remote        Add-Path
 %(families)s
 
-    Message Statistic                Sent        Received
+	Message Statistic                Sent        Received
 %(messages)s
 """.replace(
         '\t', '  '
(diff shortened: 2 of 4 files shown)


src/exabgp/reactor/api/command/rib.py
```diff
@@ -7,6 +7,8 @@
 License: 3-clause BSD. (See the COPYRIGHT file)
 """
 
+import json
```
thomas-mangin, line 10: I do not believe this import is required.

src/exabgp/bgp/neighbor.py
```diff
@@ -173,19 +173,31 @@ def clear_rib(self):
...
+            "neighbor": self.peer_address,
+            "local-ip": self.local_address if self.peer_address is not None else 'auto',
+            "local-as": self.local_as if self.local_as is not None else 'auto',
+            "peer-as": self.peer_as if self.peer_as is not None else 'auto',
+            "router-id": self.router_id,
+            "family-allowed": session
+        }
+
+        if json:
+            return data
```
thomas-mangin, line 191: This code seems to return a dict and not a string, it should be a string. While python is dynamic, I attempt to only return one type per function, like in C, to ease debugging.

src/exabgp/reactor/api/command/neighbor.py
```diff
@@ -32,38 +32,28 @@ def _pr(value):
...
-    return "disabled"
-
-
 class Neighbor(object):
     extensive_kv = '   %-20s %15s %15s %15s'
     extensive_template = """\
 Neighbor %(peer-address)s
 
-    Session                         Local
+	Session                         Local
```
thomas-mangin, line 40: It looks like spaces have been converted to tabs

src/exabgp/reactor/api/command/neighbor.py
```diff
@@ -73,99 +63,98 @@ class Neighbor(object):
...
-        formated['peer']['address'] = answer['peer-address']
-        formated['peer']['as'] = answer['peer-as']
-        formated['peer']['id'] = answer['peer-id']
-        formated['peer']['hold'] = answer['peer-hold']
-
-        return formated
-
-    @classmethod
-    def extensive(cls, answer):
+    def extensive(cls, answer, output_format='text'):
```
thomas-mangin, line 66: extensive is unlikely to be right as a @classmethod, it is a normal one class function.

src/exabgp/reactor/api/command/neighbor.py
```diff
@@ -73,99 +63,98 @@ class Neighbor(object):
     summary_template = '%-15s %-7s %9s %-12s %10d %10d'
 
     @classmethod
-    def as_dict(cls, answer):
```
thomas-mangin, line 76: as_dict should not be inlined. It would be better to have as _string added and use two functions for clarity.

src/exabgp/reactor/api/command/neighbor.py
```diff
@@ -194,22 +183,23 @@ def teardown(self, reactor, service, line):
         return False
 
 
-@Command.register('text', 'show neighbor', False, ['summary', 'extensive', 'configuration', 'json'])
+@Command.register('text', 'show neighbor', False, ['summary', 'extensive', 'configuration'])
+@Command.register('text', 'show neighbor json', False, ['summary', 'extensive'])
```
thomas-mangin, line 187: if we register <command> json, we should also register <command> text and make it work.

src/exabgp/reactor/loop.py
```diff
@@ -172,12 +172,16 @@ def neighbor(self, peer_name):
...
         return peer.neighbor
 
-    def neighbor_name(self, peer_name):
+    def neighbor_name(self, peer_name, json=False):
         peer = self._peers.get(peer_name, None)
         if not peer:
             log.critical('could not find referenced peer', 'reactor')
             return ""
-        return peer.neighbor.name()
+        if json:
```
thomas-mangin, line 180: The if statement is not required, `return peer.neighbor.name(json)` does the same thing AFAICS.

src/exabgp/reactor/loop.py
```diff
@@ -172,12 +172,16 @@ def neighbor(self, peer_name):
             return
         return peer.neighbor
 
-    def neighbor_name(self, peer_name):
+    def neighbor_name(self, peer_name, json=False):
```
thomas-mangin, line 175: json should be made mandatory and not defaulted to `False`

src/exabgp/reactor/api/command/rib.py
```diff
@@ -23,70 +25,88 @@
...
+@Command.register('text', 'show adj-rib out json', False, ['extensive',])
+@Command.register('text', 'show adj-rib in json', False, ['extensive',])
 def show_adj_rib(self, reactor, service, line):
     words = line.split()
     extensive = line.endswith(' extensive')
+    json_output = True if 'json' in words else False
     try:
         rib = words[2]
-        if not rib in ('in', 'out'):
+        if not rib in ('in', 'out', 'json'):
```
thomas-mangin, line 97: This logic seems wrong ... I would need to look more in depth to say why but I can not ATM.

thomas-mangin: For clarity it is also about the `if rib == 'json'` part too.

thomas-mangin:
Thank you for this PR, I can not accept it as it but I have no objection to adding the feature.

src/exabgp/reactor/api/command/neighbor.py
thomas-mangin: One for me to look into if the previous code was already a @classmethod

thomas-mangin: However, using an output_format as a string, not a boolean or an "enum", is wrong. String should not be given magical if it is possible to compare them.

thomas-mangin:
I will massage this patch into shape as there is too much to fix and explain.

thomas-mangin:
following the changes pushed above, it should be easier to add json support to any command.

`show adj-rib` was converted in a very inefficient way using `json.dumps`. The JSON should be generated by hand using the JSON module with a large structure as the rib allocates a lot of memory instead of using iterators, but as an example, it will do for the moment.

The JSON generated should NOT try to match the text version but instead use suitable dict structures that are easy to parse (as it is often done in the code).

__
```
@Command.register('show adj-rib out', False, ['extensive',], True)
@Command.register('show adj-rib in', False, ['extensive',], True)
def show_adj_rib(self, reactor, service, line, use_json):
```

thomas-mangin:
The patch is wrong as it mixes json payload with json formatting of answers. I will fix it.

thomas-mangin:
if you want to see what you need and I missed, please open a new PR and I will help.

PR #1245, opened 2025-02-15
debian/control: Remove unneeded debconf pre-dependency

gioele:
exabgp does not make use of debconf at all, so this (outdated) pre-dependency can be removed. An appropriate dependency will be added by `dh_installdebconf` if needed.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -35,8 +35,7 @@ Description: BGP swiss army knife of networking - Python 3 module
 
 Package: exabgp
 Architecture: all
-Pre-Depends: debconf (>= 1.5.34),
-             dpkg (>= 1.15.7.2),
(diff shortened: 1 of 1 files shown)


thomas-mangin:
@vincentbernat 
Hi Vicent, I would welcome your feedback on this PR. Also, I was wondering if I should spend some time to change the debian packaging on main to create and install a zipapp? Any thought?

thomas-mangin:
@gioele Thank you for the PR. I have not build a debian packages in years, hopefully Vincent can just ack. your patch, otherwise I will investigate.

vincentbernat:
Ack for me on this one.

If you add a `debian/` directory in the master branch, it may create conflicts and make it more difficult to work in debian/sid. But this is a minor inconvenience.

PR #1255, opened 2025-06-27
Implement RFC 9514

Quentique:
https://datatracker.ietf.org/doc/rfc9514/

Add SRv6 support to BGP-LS.

Tested between Exabgp <--> Cisco IOS XRv


thomas-mangin:
I would suggest a few modifications as another patch: 
1. please retain the copyright under your own name, we do not ask for copyright assignment 
2. the code uses json for parsing, I am not sure why, as I have not looked at the RFC in depth, I am not sure if the code could/should be implemented differently.
3. the generated json does not add any space after { in some cases (horror but I do not have OCD so it can remain as if required).

Otherwise LGTM - merging.

thomas-mangin:
Thank you very much for taking the time to add this feature.

PR #1256, opened 2025-06-27
github: use latest Ubuntu image for old Python versions

vincentbernat:
This should not prevent to use older Python versions as actions/setup-python is able to download older Python versions from actions/python-versions.

### DIFF
--- a/.github/workflows/functional-3.6.yml
+++ b/.github/workflows/functional-3.6.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.6"]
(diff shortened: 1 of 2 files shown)


vincentbernat:
My bad, I understand that the Python interpreters provided by actions/python-versions are not available for all versions of Ubuntu. Maybe, just deprecated Python 3.6 (end of maintenance in 2021, not available in any supported Debian, available in Ubuntu 18.04)?

thomas-mangin:
Yes, I will end support for 3.6 as there is no easy way to run the CI anymore.

PR #1263, opened 2025-09-01
Implement ip-ifname argument to setup ip addresses on physical interfaces

david-guti:
On certain deployments, hosts could use anycast addresses (/32 and /128) inside a routable network (/24 and /64) so that, even if the BGP cloud established via exabgp was down, traffic would continue to be routed to the destinations.

To enable this feature, while setting up the '--ip' addresses via the healthcheck script, healthcheck.py needs to be able to setup '--ip' on physical interfaces. One of the reasons is that the linux ipv6 implementation doesn't allow traffic bridging between physical and loopback interfaces.

The proposed PR implements a '--ip-ifname <ip_address>%<phy_ifname>' argument to select the physical NIC where the '--ip' addresses need to be setup.

### DIFF
--- a/src/exabgp/application/healthcheck.py
+++ b/src/exabgp/application/healthcheck.py
@@ -78,8 +78,8 @@ def setargs(parser):
     parser.add_argument("--name", "-n", metavar="NAME", help="name for this healthchecker")
     parser.add_argument("--config", "-F", metavar="FILE", type=open, help="read configuration from a file")
     parser.add_argument("--pid", "-p", metavar="FILE", type=argparse.FileType('w'), help="write PID to the provided file")
-    parser.add_argument("--user", metavar="USER", help="set user after setting loopback addresses")
-    parser.add_argument("--group", metavar="GROUP", help="set group after setting loopback addresses")
+    parser.add_argument("--user", metavar="USER", help="set user after setting ip addresses")
+    parser.add_argument("--group", metavar="GROUP", help="set group after setting ip addresses")
 
     g = parser.add_argument_group("checking healthiness")
     g.add_argument("--interval", "-i", metavar='N', default=5, type=float, help="wait N seconds between each healthcheck (zero to exit after first announcement)")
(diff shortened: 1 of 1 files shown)


thomas-mangin:
My advice for anycast is to bring the IP on the loopback interface as you the do not have any MAC pollution on the LAN. It require to change the kernel networking settings to not filter packet for that IP coming from another interface.

/etc/sysctl.conf
```
net.ipv4.conf.all.arp_filter = 1
net.ipv4.conf.all.arp_ignore = 1
net.ipv4.conf.eth0.arp_ignore = 1
net.ipv4.conf.all.arp_announce = 2
net.ipv4.conf.eth0.arp_announce = 2
```

But this feature may be useful if people have reasons - while I can not see one - to bring the IP on the LAN/WAN.

@vincentbernat are you ok with it?

david-guti:
Thank you Thomas for your comments.

Not sure I understand your reference to the MAC pollution. From the LAN point of view, an anycast address on the physical NIC (i.e. eth0) will look like a virtual IP floating between high availability nodes in a cluster. 

Note that the physical ip address of the host (i.e. eth0) and the anycast need to be on the same subnet, otherwise this extension doesn't bring anything.

Given that the IP anycast address will be announced through the 'next-hop', which is the ip address of the physical NIC, I'm not sure when traffic would come from a different interface (?).

thomas-mangin:
Pollution is not the best word. Yes, it bring the IP on the LAN segment when it is not required if placed on the loopback.

Normally I bring service IP for anycast as /32 with the next-hop as the “normal” IP of the server on the Layer 2 used to bring BGP

david-guti:
Yes, it's the same schema we are using. 

Just FYI, see an example running, with two 'lo' and two 'eth0' anycast addresses.

1: lo    inet **10.142.17.15/32** scope global lo\       valid_lft forever preferred_lft forever
1: lo    inet6 **fd01:1458:301:1100::100:f/128** scope global \       valid_lft forever preferred_lft forever
2: eth0    inet 10.142.16.13/27 brd 128.142.16.31 scope global noprefixroute eth0\       valid_lft forever preferred_lft forever
2: eth0    inet **10.142.16.15/32** scope global eth0\       valid_lft forever preferred_lft forever
2: eth0    inet6 fd01:1458:301:1000::100:d/64 scope global noprefixroute \       valid_lft forever preferred_lft forever
2: eth0    inet6 **fd01:1458:301:1000::100:f/128** scope global \       valid_lft forever preferred_lft forever

On the routers, arp looks like:

7a:b2:45:9c:a0:1a 10.142.16.13   irb.33 [ae216.0]         permanent remote
7a:b2:45:9c:a0:1a 10.142.16.15   irb.33 [ae216.0]         permanent remote

And nothing new on exabgp:

[root@host-513 ~]# exabgpcli show adj-rib out
neighbor 10.24.48.31 ipv4 unicast 10.142.16.15/32 next-hop 10.142.16.13
neighbor 10.24.48.31 ipv4 unicast 10.142.17.15/32 next-hop 10.142.16.13
neighbor fd01:1458:300:a::31 ipv6 unicast fd01:1458:301:1000::100:f/128 next-hop fd01:1458:301:1000::100:d
neighbor fd01:1458:300:a::31 ipv6 unicast fd01:1458:301:1100::100:f/128 next-hop fd01:1458:301:1000::100:d

thomas-mangin:
@david-guti can you please fix https://github.com/Exa-Networks/exabgp/actions/runs/17371503911/job/49479315973?pr=1263 and https://github.com/Exa-Networks/exabgp/actions/runs/17371503915/job/49479315951?pr=1263

david-guti:
Should look better now.

david-guti:
Great! Thank you Thomas. This merge will simplify our service operation!

src/exabgp/application/healthcheck.py
```diff
@@ -557,12 +590,11 @@ def main():
         options.pid.close()
     try:
         # Setup IP to use
-        options.ips = options.ips or loopback_ips(options.label, False, options.label_exact_match)
```
shadow-protocol, line 560: Doesn't this change default behaviour (ie. no `--ip` given) from picking up `lo` addresses, to _not_ picking up any? If so, how does that relate to the spirit of the rest of this merge? I'm wondering if this line was unintentional as it wasn't mentioned in the commit msg.

Disclaimer: read-only experience. Read @vincentbernat 3tier blog post, then their lab setup script, then HEAD on healthcheck here and got pretty confused about how the service VIP ends up getting announced. This line change explains my confusion... I think. Grain of salt and all that...

david-guti: I think I can modify the code to continue to behave in an equivalent manner only for loopback interfaces. This would ensure backwards compatibility.

PR #1267, opened 2025-09-23
Add Github action to build and publish container

jschlyter:
First cut at building an "official" ExaBGP container.

- Allow all Python 3.13.x versions
- Use slim container for runtime (save about 50% space)
- Build containers for both `linux/amd64` and `linux/arm64`, others platforms can be added easily


thomas-mangin:
Thank you

PR #1374, opened 2026-04-04
chore: remove some cruft files

vincentbernat:
I don't think they were expected to be committed in the repository.


thomas-mangin:
Thank you @vincentbernat - I dropped you an email

PR #1375, opened 2026-04-08
Immediately return empty Capabilties if length is 0

brijohn:
Receiving an open message with no capabilities was generating an exception due not immediately returning an empty set of capabilities if the capabilities length was 0 in the capability unpack function.

### DIFF
--- a/src/exabgp/bgp/message/open/capability/capabilities.py
+++ b/src/exabgp/bgp/message/open/capability/capabilities.py
@@ -246,6 +246,10 @@ class Capabilities(dict):
 
         # Extended optional parameters
         option_len = data[0]
+
+        if not option_len:
+            return capabilities
+
         option_type = data[1]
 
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Hi @brijohn

Thank you for catching the OPEN capabilities problem. Your fix was correct, I did not merge the PR as-is as I got in a slight variation with a bit of extra cleanup.

As it is a pretty bad issue, I released 5.0.8 (commit 60f0a90b), with the commit message and changelog both crediting your PR as the alternative.

There has been a "small run" of fixes in this code recently, the send side in 5.0.7 and now the receive side from your report in 5.0.8, and I am genuinely crossing fingers that this one stays put for a while.

Thanks again for taking the time to investigate it and provide a good working solution.

Thomas

brijohn:
@thomas-mangin Thanks for getting a version of this in. 

Yes we had also found the issues with sending 0 length capabilities such as RouteRefresh and always sending the Hostname even when one was not set, but those had already been fixed in 5.0.7 so that's a couple less i need to submit.

PR #1377, opened 2026-04-09
A hostname that starts or ends with a digit should be considered valid.

brijohn:
Hostnames should be able to start with a digit as well as end with one. 

We use several VMS for testing that end in numbers indicating the VM that is being accessed, and noticed that anytime we set the hostname on them exabgp would fail while parsing the hostname field in the config.


thomas-mangin:
Thank you @brijohn as I am trying to get version 5 stable. I will then do the same for version 6 any further effort will only be on https://github.com/ze-software/ze

I will forward port this change

thomas-mangin:
Forward-ported to main in commit 72e4c39a4.

PR #1380, opened 2026-04-10
Add IPV6_EXTENDED_COMMUNITY to Attributes.representation

brijohn:
The IPV6_EXTENDED_COMMUNITY attribute code is currently missing the representation dictionary. This causes it to displayed using the "attribute-0x{:02X}-0x{:02X}" format, instead of using a more human readable name.


thomas-mangin:
Thank you, forward ported to main.

PR #1385, opened 2026-04-21
Rework the show adj-rib out json output

brijohn:
This commit is a bit more extensive but there are currently several broken issues with the json output for the "show adj-rib" command. 

The first is that it doesn't work when trying to uses the extensive modifier the whole command throws an exception and errors out. 

Secondly they way it is currently written it only works it will only work when all your routes are inet based due to its use of calling cider.prefix() on each nlri.

There were also some issues when changing the api.chunk setting causing it to return the same routes multiple times.

This command had several issues with it that made it not work.
   * Extensive option was throwing an exception when used.
   * Listing the routes only worked for inet nrlis.
   * Splitting json.dumps at newline is not needed, as it only outputs a single line.
   * No longer include an array of routes, this was not really behaving as intended with api.chuncks set to any value higher then 1 and in that case the array only ever had one value anyways.

### DIFF
--- a/src/exabgp/bgp/message/update/attribute/aspath.py
+++ b/src/exabgp/bgp/message/update/attribute/aspath.py
@@ -162,6 +162,15 @@ class ASPath(Attribute):
         self._json = json.dumps(jason)
         return self._json
 
+    def as_dict(self):
+        result = {}
+        for pos, content in enumerate(self.aspath):
+            result[pos] = {
+                'element': content.NAME,
+                'value': list(content),
+            }
+        return result
+
     @classmethod
     def _new_aspaths(cls, data, asn4, klass=None):
         backup = data
--- a/src/exabgp/bgp/message/update/attribute/attributes.py
+++ b/src/exabgp/bgp/message/update/attribute/attributes.py
@@ -144,6 +144,33 @@ class Attributes(dict):
             else:
                 yield ' {} {}'.format(name, presentation % str(attribute))
 
+    def _generate_dict(self):
+        for code in sorted(self.keys()):
+            if code in Attributes.NO_GENERATION:
+                continue
+
+            attribute = self[code]
+
+            if code not in self.representation:
+                key = f"attribute-0x{code:02X}-0x{attribute.FLAG:02X}"
+                yield key, str(attribute)
+                continue
+
+            how, _, name, _, presentation = self.representation[code]
+            if how == 'boolean':
+                yield name, self.has(code)
+            elif how == 'string':
+                yield name, str(attribute)
+            elif how == 'list':
+                yield name, attribute.as_dict()
+            elif how == 'integer':
+                yield name, int(str(attribute))
+            elif how == 'inet':
+                yield name, str(attribute)
+            # Should never be ran
+            else:
+                yield name, str(attribute)
+
     def _generate_json(self):
         for code in sorted(self.keys()):
             # remove the next-hop from the attribute as it is define with the NLRI
@@ -180,6 +207,7 @@ class Attributes(dict):
         self._str = ''
         self._idx = ''
         self._json = ''
+        self._dict = {}
         # The parsed attributes have no mp routes and/or those are last
         self.cacheable = True
 
@@ -201,6 +229,7 @@ class Attributes(dict):
 
             self._str = ''
             self._json = ''
+            self._dict = {}
 
             for community in attribute.communities:
                 self[attribute.ID].add(community)
@@ -208,6 +237,8 @@ class Attributes(dict):
 
         self._str = ''
         self._json = ''
+        self._dict = {}
+
         self[attribute.ID] = attribute
 
     def remove(self, attrid):
@@ -267,6 +298,11 @@ class Attributes(dict):
             message += attribute.pack(negotiated)
 
         return message
+
+    def as_dict(self):
+        if not self._dict:
+            self._dict = dict(self._generate_dict())
+        return self._dict
 
     def json(self):
         if not self._json:
--- a/src/exabgp/bgp/message/update/attribute/bgpls/link/sradj.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/link/sradj.py
@@ -85,3 +85,13 @@ class SrAdjacency(FlagLS):
                 'undecoded-sids': self.undecoded,
             },
         )
+
+    def as_dict(self):
+        return {
+            "sr-adj": {
+                "flags": self.flags,
+                "sids": self.sids,
+                "weight": self.weight,
+                "undecoded-sids": self.undecoded,
+            }
+        }
--- a/src/exabgp/bgp/message/update/attribute/bgpls/link/sradjlan.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/link/sradjlan.py
@@ -89,5 +89,8 @@ class SrAdjacencyLan(FlagLS):
     def json(self, compact=None):
         return f'"sr-adj-lan-sids": {json.dumps(self.sr_adj_lan_sids)}'
 
+    def as_dict(self):
+        return {"sr-adj-lan-sids": self.sr_adj_lan_sids}
+
     def merge(self, klass):
         self.sr_adj_lan_sids.extend(klass.sr_adj_lan_sids)
--- a/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6capabilities.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6capabilities.py
@@ -58,3 +58,6 @@ class Srv6Capabilities(BaseLS):
             },
             indent=compact,
         )
+
+    def as_dict(self):
+        return {"srv6-capabilities": {"flags": self.flags}}
--- a/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6endpointbehavior.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6endpointbehavior.py
@@ -62,3 +62,12 @@ class Srv6EndpointBehavior(BaseLS):
             },
             indent=compact,
         )
+
+    def as_dict(self):
+        return {
+            "srv6-endpoint-behavior": {
+                "endpoint-behavior": self.endpoint_behavior,
+                "flags": self.flags,
+                "algorithm": self.algorithm,
+            }
+        }
--- a/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6endx.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6endx.py
@@ -102,3 +102,6 @@ class Srv6EndX(FlagLS):
 
     def json(self, compact=None):
         return '"srv6-endx": [ {} ]'.format(', '.join([json.dumps(d, indent=compact) for d in self.content]))
+
+    def as_dict(self):
+        return {"srv6-endx": self.content}
--- a/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6lanendx.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6lanendx.py
@@ -124,6 +124,9 @@ class Srv6LanEndXISIS(FlagLS):
     def json(self, compact=None):
         return '"srv6-lan-endx-isis": [ {} ]'.format(', '.join([json.dumps(d, indent=compact) for d in self.content]))
 
+    def as_dict(self):
+        return {"srv6-lan-endx-isis": self.content}
+
 
 @LinkState.register()
 class Srv6LanEndXOSPF(FlagLS):
@@ -162,3 +165,6 @@ class Srv6LanEndXOSPF(FlagLS):
 
     def json(self, compact=None):
         return '"srv6-lan-endx-ospf": [ {} ]'.format(', '.join([json.dumps(d, indent=compact) for d in self.content]))
+
+    def as_dict(self):
+        return {"srv6-lan-endx-ospf": self.content}
--- a/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6locator.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6locator.py
@@ -61,3 +61,12 @@ class Srv6Locator(FlagLS):
             },
             indent=compact,
         )
+
+    def as_dict(self):
+        return {
+            "srv6-locator": {
+                "flags": self.flags,
+                "algorithm": self.algorithm,
+                "metric": self.metric,
+            }
+        }
--- a/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6sidstructure.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/link/srv6sidstructure.py
@@ -70,3 +70,13 @@ class Srv6SidStructure(BaseLS):
             },
             indent=compact,
         )
+
+    def as_dict(self):
+        return {
+            "srv6-sid-structure": {
+                "loc_block_len": self.loc_block_len,
+                "loc_node_len": self.loc_node_len,
+                "func_len": self.func_len,
+                "arg_len": self.arg_len,
+            }
+        }
--- a/src/exabgp/bgp/message/update/attribute/bgpls/linkstate.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/linkstate.py
@@ -93,6 +93,12 @@ class LinkState(Attribute):
         content = ', '.join(d.json() for d in self.ls_attrs)
         return f'{{ {content} }}'
 
+    def as_dict(self):
+        result = {}
+        for d in self.ls_attrs:
+            result.update(d.as_dict())
+        return result
+
     def __str__(self):
         return ', '.join(str(d) for d in self.ls_attrs)
 
@@ -115,6 +121,11 @@ class BaseLS:
         except TypeError:
             # not a basic type
             return f'"{self.JSON}": "{self.content.decode("utf-8")}"'
+
+    def as_dict(self):
+        if isinstance(self.content, bytes):
+            return {self.JSON: self.content.decode("utf-8")}
+        return {self.JSON: self.content}
 
     def __repr__(self):
         return '{}: {}'.format(self.REPR, self.content)
@@ -153,6 +164,9 @@ class GenericLSID(BaseLS):
         merged = ', '.join([f'"{hexstring(_)}"' for _ in self.content])
         return f'"generic-lsid-{self.TLV}": [{merged}]'
 
+    def as_dict(self):
+        return {f"generic-lsid-{self.TLV}": [hexstring(c) for c in self.content]}
+
     @classmethod
     def unpack(cls, data):
         return cls(data)
@@ -167,6 +181,9 @@ class FlagLS(BaseLS):
 
     def json(self, compact=None):
         return f'"{self.JSON}": {json.dumps(self.flags)}'
+
+    def as_dict(self):
+        return {self.JSON: self.flags}
 
     @classmethod
     def unpack_flags(cls, data):
--- a/src/exabgp/bgp/message/update/attribute/bgpls/node/srcap.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/node/srcap.py
@@ -98,3 +98,8 @@ class SrCapabilities(FlagLS):
 
     def json(self, compact=None):
         return f'{FlagLS.json(self)}, "sids": {self.sids}'
+
+    def as_dict(self):
+        result = FlagLS.as_dict(self)
+        result["sids"] = self.sids
+        return result
--- a/src/exabgp/bgp/message/update/attribute/bgpls/prefix/srprefix.py
+++ b/src/exabgp/bgp/message/update/attribute/bgpls/prefix/srprefix.py
@@ -86,3 +86,11 @@ class SrPrefix(FlagLS):
 
     def json(self, compact=None):
         return f'"sr-prefix-flags": {json.dumps(self.flags)}, "sids": {json.dumps(self.sids)}, "undecoded-sids": {json.dumps(self.undecoded)}, "sr-algorithm": {json.dumps(self.sr_algo)}'
+
+    def as_dict(self):
+        return {
+            "sr-prefix-flags": self.flags,
+            "sids": self.sids,
+            "undecoded-sids": self.undecoded,
+            "sr-algorithm": self.sr_algo,
+        }
--- a/src/exabgp/bgp/message/update/attribute/clusterlist.py
+++ b/src/exabgp/bgp/message/update/attribute/clusterlist.py
@@ -52,6 +52,9 @@ class ClusterList(Attribute):
     def json(self):
         return '[ {} ]'.format(', '.join(['"{}"'.format(str(_)) for _ in self.clusters]))
 
+    def as_dict(self):
+        return [str(_) for _ in self.clusters]
+
     @classmethod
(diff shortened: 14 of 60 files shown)


thomas-mangin:
Thank you @brijohn

thomas-mangin:
ah! problem, this is a commit for 5.0 - how does it affect backward compatibility?

thomas-mangin:
5.0 `as_dict()` patch vs Main `json()` representation comparison

Comparison of 5.0 commit `976e69340` ("Rework the show adj-rib out json output") against `main`.

## Top-level JSON structure (rib.py)

**5.0 after patch:**
```json
{"peer": {"address": "1.2.3.4"}, "route": {"nlri": {...}, "attributes": {...}}}
```

**Main current:**
```json
{"1.2.3.4": {"routes": [{"prefix": "10.0.0.0/24", "family": "ipv4 unicast"}]}}
```

Completely different structure. 5.0 emits one JSON object per route with full NLRI + attributes. Main emits a growing array with only prefix + family (no attributes at all).

## INET NLRI

**5.0 `as_dict()`:**
```json
{"prefix": "10.0.0.0/24", "nexthop": "1.2.3.4", "family": {"afi": "ipv4", "safi": "unicast"}}
```

**Main `json()`:**
```json
{"nlri": "10.0.0.0/24"}
```

- `as_dict()` includes `nexthop` and `family`; `json()` does not
- `as_dict()` uses `"prefix"` key; `json()` uses `"nlri"` key
- `as_dict()` family is structured `{"afi": "...", "safi": "..."}`; `json()` has no family

## EVPN base (unparsed)

**5.0 `as_dict()`:**
```json
{"code": 2, "parsed": false, "raw": "0102...", "name": "MAC", "family": {"afi": "l2vpn", "safi": "evpn"}}
```

**Main `json()`:**
```json
{"code": 2, "parsed": false, "raw": "0102..."}
```

`as_dict()` adds `"name"` and `"family"` fields.

## EVPN MAC (parsed)

**5.0 `as_dict()`:**
```json
{"code": 2, "parsed": true, "name": "MAC", "family": {...}, "rd": "1:1", "esi": "...", "ethernet-tag": 0, "mac": "aa:bb:cc:dd:ee:ff", "maclen": 48, "label": [[100]], "ip": null}
```

**Main `json()`:**
```json
{"code": 2, "parsed": true, "raw": "...", "name": "MAC", "rd": "1:1", "esi": "...", "ethernet-tag": {...}, "mac": {...}, "label": "..."}
```

- `as_dict()`: `"ethernet-tag"` is an integer; `json()` calls `self.etag.json()` producing a JSON object string
- `as_dict()`: `"mac"` is a plain string; `json()` calls `self.mac.json()` which includes compact formatting
- `as_dict()`: `"label"` is `[[value, raw?]]` array; `json()` calls `self.label.json()` which is a string
- `as_dict()` has `"maclen"` field; `json()` does not
- `as_dict()` has `"family"` field; `json()` does not

## Flow NLRI

**5.0 `as_dict()`:**
```json
{"rules": {"source": ["10.0.0.0/24"], "destination": ["20.0.0.0/24"]}, "nexthop": "1.2.3.4", "rd": null, "family": {"afi": "ipv4", "safi": "flow"}}
```

**Main `json()`:**
```json
{"source": ["10.0.0.0/24"], "destination": ["20.0.0.0/24"], "rd": "...", "string": "flow ..."}
```

- `as_dict()` nests rules under `"rules"` key; `json()` puts rules at top level
- `as_dict()` has `"nexthop"` and `"family"`; `json()` has `"string"` (human-readable)
- `as_dict()` omits the compatibility `"string"` field

## VPLS NLRI

**5.0 `as_dict()`:**
```json
{"rd": "1:1", "endpoint": 1, "base": 100, "offset": 0, "size": 10, "nexthop": "1.2.3.4", "family": {"afi": "l2vpn", "safi": "vpls"}}
```

**Main `json()`:**
```json
{"rd": "1:1", "endpoint": 1, "base": 100, "offset": 0, "size": 10}
```

`as_dict()` adds `"nexthop"` and `"family"`.

## ASPath attribute

5.0 `as_dict()` and main `json()` are **identical in structure:**
```json
{"0": {"element": "SEQUENCE", "value": [65000, 65001]}}
```

## Communities / ClusterList attributes

**Functionally equivalent** when serialized. Both produce arrays of strings.

## Label handling

**5.0 `as_dict()`:** `"label": [[100], [200, "0x00C801"]]` (nested arrays with optional raw)

**Main `json()`:** Calls `self.label.json()` which returns `"label": [100]` (simpler format)

Different structure for multi-label / raw label cases.

## Attributes container (`_generate_dict` vs `_generate_json`)

| Category | 5.0 `_generate_dict` | Main `_generate_json` |
|----------|---------------------|----------------------|
| `'multiple'` type | **Missing**, falls to `else` branch | Has dedicated handler with per-name iteration |
| `'boolean'` | Returns Python `True/False` | Returns `"true"/"false"` strings |
| `'integer'` | Returns `int(str(attribute))` | Returns `%s` format (string) |
| `'list'` | Calls `attribute.as_dict()` | Calls `attribute.json()` |
| NEXT_HOP skip | Skips via `NO_GENERATION` | Conditionally includes via `include_nexthop` param |
| Generic attrs | `"attribute-0x{code}-0x{flag}": str(attr)` | Same pattern but as JSON string |

## BGP-LS, SR, SRv6 attributes

Structurally equivalent between `as_dict()` and `json()` (same keys, same values), just dict vs JSON string.

## Missing `as_dict()` methods (5.0 patch is incomplete)

These `'list'` type attributes will **crash** when `_generate_dict` calls `.as_dict()`:

- `ExtendedCommunities` (code 16)
- `ExtendedCommunitiesIPv6` (code 25)
- `LargeCommunities` (code 32)

## Summary

| Category | Severity | Description |
|----------|----------|-------------|
| Top-level JSON structure | Breaking | Completely different: `{peer:{}, route:{nlri:{}, attributes:{}}}` vs `{ip: {routes: [{prefix, family}]}}` |
| NLRI key naming | Breaking | `"prefix"` vs `"nlri"` for INET |
| nexthop/family inclusion | Additive | `as_dict()` adds `nexthop` + `family` everywhere; `json()` omits them in NLRI |
| Flow rules nesting | Breaking | Rules nested under `"rules"` key vs flat |
| EVPN sub-fields | Different | `as_dict()` unwraps sub-objects to primitives; `json()` delegates to `.json()` methods |
| Label format | Different | `[[value, raw?]]` vs simple format |
| `'multiple'` handler | Missing | 5.0 `_generate_dict` has no `'multiple'` case (extended communities) |
| Missing `as_dict()` | Crash | ExtendedCommunities, LargeCommunities, ExtendedCommunitiesIPv6 have no `as_dict()` |

thomas-mangin:
5.0 `as_dict()` patch: before vs after representation comparison

Comparison of 5.0 commit `976e69340` ("Rework the show adj-rib out json output").
This document compares the **existing `json()` output** (before the patch) against the **new `as_dict()` output** (after the patch), within 5.0 itself.

A separate section at the end notes where main diverges from both.

---

## Top-level JSON structure (rib.py)

### Before (5.0 `to_json`)
```json
{"1.2.3.4": {"routes": [{"prefix": "10.0.0.0/24", "family": "ipv4 unicast"}]}}
```
- Keyed by neighbor IP
- Routes accumulated in an array
- Only `prefix` + `family` per route (no attributes)
- `json.dumps().split('\n')` (pointless, dumps produces one line)
- Broken: `nlri.cidr.prefix()` only works for INET

### After (5.0 `to_json` rewritten)
```json
{"peer": {"address": "1.2.3.4"}, "route": {"nlri": {...}, "attributes": {...}}}
```
- Flat structure with `"peer"` and `"route"` keys
- One JSON object emitted per route
- Full NLRI + all attributes included
- Works for all address families

**Verdict: completely different schema, breaking change.**

---

## INET NLRI

### Before (`json()`)
```json
{"nlri": "10.0.0.0/24"}
```
- Key is `"nlri"`
- No nexthop, no family
- Optional `"path-information"` if addpath

### After (`as_dict()`)
```json
{"prefix": "10.0.0.0/24", "nexthop": "1.2.3.4", "family": {"afi": "ipv4", "safi": "unicast"}}
```
- Key renamed to `"prefix"`
- Adds `"nexthop"` and structured `"family"`
- No path-information field

**Differences:**
- `"nlri"` renamed to `"prefix"`
- `nexthop` and `family` added
- `path-information` dropped

---

## Label NLRI (INET + MPLS labels)

### Before (`json()`)
No `json()` override on Label. Inherits `INET.json()` which shows the NLRI key only. Labels are not part of the NLRI json output.

### After (`as_dict()`)
```json
{"prefix": "10.0.0.0/24", "nexthop": "1.2.3.4", "family": {...}, "label": [[100], [200, 524289]]}
```
- Inherits from `INET.as_dict()`
- Adds `"label"` as array of `[value]` or `[value, raw]` pairs

**Difference: labels now visible in NLRI output (new field).**

---

## IPVPN NLRI (Label + RD)

### Before (`json()`)
No `json()` override. Inherits from Label/INET.

### After (`as_dict()`)
```json
{"prefix": "10.0.0.0/24", "nexthop": "...", "family": {...}, "label": [...], "rd": "1:1"}
```
- Inherits from `Label.as_dict()`
- Adds `"rd"` field

**Difference: RD now visible in NLRI output (new field).**

---

## Flow NLRI

### Before (`json()`)
```json
{"source": ["10.0.0.0/24"], "destination": ["20.0.0.0/24"], "rd": "...", "next-hop": "1.2.3.4", "string": "flow source 10.0.0.0/24 ..."}
```
- Rules at top level
- `"next-hop"` key (with hyphen)
- `"string"` compatibility field with human-readable form
- RD via `self.rd.json()` which outputs `"rd": "..."`

### After (`as_dict()`)
```json
{"rules": {"source": ["10.0.0.0/24"], "destination": ["20.0.0.0/24"]}, "nexthop": "1.2.3.4", "rd": "1:1", "family": {"afi": "ipv4", "safi": "flow"}}
```
- Rules nested under `"rules"` key
- `"nexthop"` key (no hyphen)
- No `"string"` compatibility field
- `"family"` added
- RD via `self.rd._str()` (plain string, not JSON fragment)

**Differences:**
- Rules nesting (`flat` vs `"rules": {...}`)
- `"next-hop"` renamed to `"nexthop"`
- `"string"` field removed
- `"family"` field added
- AND operator handling differs in rule formatting

---

## VPLS NLRI

### Before (`json()`)
```json
{"rd": "1:1", "endpoint": 1, "base": 100, "offset": 0, "size": 10}
```
- RD via `self.rd.json()` which outputs `"rd": "1:1"` as JSON fragment

### After (`as_dict()`)
```json
{"rd": "1:1", "endpoint": 1, "base": 100, "offset": 0, "size": 10, "nexthop": "1.2.3.4", "family": {"afi": "l2vpn", "safi": "vpls"}}
```
- RD via `self.rd._str()` (same value, different code path)
- `"nexthop"` and `"family"` added

**Difference: additive only (nexthop + family).**

---

## RTC NLRI

### Before (`json()`)
No `json()` method on RTC. Would use `NLRI` base which likely produces minimal output.

### After (`as_dict()`)
```json
{"origin": 65000, "rt": "65000:100", "nexthop": "1.2.3.4", "family": {"afi": "ipv4", "safi": "rtc"}}
```

**Difference: entirely new representation (no prior json).**

---

## EVPN base (unparsed / generic)

### Before (`json()`)
```json
{"code": 2, "parsed": false, "raw": "0102..."}
```

### After (`as_dict()`)
```json
{"code": 2, "parsed": false, "raw": "0102...", "name": "MAC", "family": {"afi": "l2vpn", "safi": "evpn"}}
```

**Difference: `"name"` and `"family"` fields added.**

---

## EVPN MAC

### Before (`json()`)
```json
{"code": 2, "parsed": true, "raw": "...", "name": "MAC", "rd": "1:1", "esi": "...", "ethernet-tag": 0, "mac": "aa:bb:cc:dd:ee:ff", "label": [[100, 524289]], "ip": "10.0.0.1"}
```
- Sub-objects rendered via their own `.json()` methods:
  - `self.rd.json()` produces `"rd": "1:1"` (JSON key-value fragment)
  - `self.esi.json()` produces `"esi": "..."` (JSON key-value fragment)
  - `self.etag.json()` produces `"ethernet-tag": 0` (JSON key-value fragment)
  - `self.mac.json()` produces `"mac": "..."` (JSON key-value fragment)
  - `self.label.json()` produces `"label": [ [100, 524289] ]`

### After (`as_dict()`)
```json
{"code": 2, "parsed": true, "name": "MAC", "family": {...}, "nexthop": null, "rd": "1:1", "esi": "...", "ethernet-tag": 0, "mac": "aa:bb:cc:dd:ee:ff", "maclen": 48, "label": [[100], [100, 524289]], "ip": null}
```
- Sub-objects unwrapped manually:
  - `self.rd._str()` or `None` if NORD
  - `str(self.esi)` 
  - `self.etag.tag` (integer directly)
  - `str(self.mac)`
  - Label built manually as `[[value]` or `[value, raw]]`

**Differences:**
- `"raw"` field present in json(), absent in as_dict()
- `"nexthop"` added (always present, null if none)
- `"family"` added
- `"maclen"` added (new field)
- `"ip"` always present in as_dict() (null if none); json() only includes when truthy
- `"rd"` is `None` for NORD in as_dict(); json() uses `self.rd.json()` which returns empty string for no RD
- Label format: json() delegates to `label.json()` producing `[ [value, raw] ]`; as_dict() builds `[[value]]` or `[[value, raw]]`

---

## EVPN EthernetAD

Same pattern as EVPN MAC. Key differences from its own `json()`:
- `"raw"` dropped
- `"nexthop"` added
- `"family"` added (via base EVPN.as_dict())
- `"esi"` via `str()` instead of `.json()` fragment
- `"ethernet-tag"` is `self.etag.tag` (int) instead of `self.etag.json()` fragment
- Label built manually instead of `self.label.json()`

---

## EVPN Multicast

Same pattern. Additionally:
- `"ip"` always present (null if none) vs only included when truthy in json()

---

## EVPN Prefix (Type 5)

Same pattern. Additionally:
- `"gateway"` and `"iplen"` fields present in both json() and as_dict()
- `"ip"` always present in as_dict() vs always present in json() (same here)

---

## EVPN EthernetSegment

Same pattern. Additionally:
- `"ip"` always present (null if none) in as_dict() vs always present in json()

---

## ASPath attribute

### Before (`json()`)
```python
json.dumps({0: {"element": "SEQUENCE", "value": [65000, 65001]}})
```

### After (`as_dict()`)
```python
{0: {"element": "SEQUENCE", "value": [65000, 65001]}}
```

**Identical structure. Only difference: dict vs JSON string.**

---

## ClusterList attribute

### Before (`json()`)
```
[ "1.1.1.1", "2.2.2.2" ]
```
(JSON string)

### After (`as_dict()`)
```python
["1.1.1.1", "2.2.2.2"]
```
(Python list)

**Identical structure.**

---

## Communities attribute

### Before (`json()`)
```
[ "65000:100", "65000:200" ]
```
(JSON string, delegates to each `community.json()`)

### After (`as_dict()`)
```python
["65000:100", "65000:200"]
```
(Python list, uses `str(community)`)

**Potential difference:** `community.json()` and `str(community)` may produce different output for some community types (e.g., well-known communities like NO_EXPORT). Not verified.

---

## LinkState attribute

### Before (`json()`)
```
{ "igp-metric": 10, "sr-prefix-flags": {...}, ... }
```
(JSON string, joins sub-attribute `.json()` calls with commas)

### After (`as_dict()`)
```python
{"igp-metric": 10, "sr-prefix-flags": {...}, ...}
```
(Python dict, merges sub-attribute `.as_dict()` calls via `.update()`)

**Same structure. But `as_dict()` uses `dict.update()` which silently overwrites duplicate keys if two sub-attributes produce the same key. `json()` would produce duplicate keys in the JSON string (technically invalid JSON but not caught).**

---

## PrefixSid attribute

Same pattern as LinkState. Merges sub-TLV `.as_dict()` results.

---

## SRv6 SID Information

### Before (`json()`)
```json
{"sid": "2001:db8::1", "flags": 0, "endpoint_behavior": 6}
```
(JSON string, uses `%s` formatting with `self.sid` and `%d` with `self.behavior`)

### After (`as_dict()`)
```json
{"sid": "2001:db8::1", "flags": 0, "endpoint_behavior": 6}
```

**Identical structure.** Both include sub-sub-TLVs via iteration.

---

## SRv6 SID Structure

### Before (`json()`)
```json
{"structure": {"locator-block-length": 40, "locator-node-length": 24, ...}}
```

### After (`as_dict()`)
```json
{"structure": {"locator-block-length": 40, "locator-node-length": 24, ...}}
```

**Identical structure.**

---

## BGP-LS SRv6 SID Structure (link/srv6sidstructure.py)

### Before (`json()`)
Keys: `"loc_block_len"`, `"loc_node_len"`, `"func_len"`, `"arg_len"`

### After (`as_dict()`)
Keys: `"loc_block_len"`, `"loc_node_len"`, `"func_len"`, `"arg_len"`

**Identical.** (Note: these use underscore keys unlike the SR attribute version which uses hyphenated keys. This inconsistency exists in both json() and as_dict().)

---

## Attributes container (`_generate_dict` vs `_generate_json`)

### Before (`_generate_json`)
```python
elif how == 'multiple':
    for n in name:
        value = attribute.json(n)
        if value:
            yield '"{}": {}'.format(n, presentation % value)
```

### After (`_generate_dict`)
```python
# 'multiple' case is MISSING
# Falls through to else: yield name, str(attribute)
```

**The `'multiple'` type handler is missing in `_generate_dict`.** In the 5.0 representation dict, no attribute currently uses `'multiple'`, so this doesn't crash today. But main has extended community attributes that could use it.

### Other differences in the container:

| Type | `_generate_json` (before) | `_generate_dict` (after) |
|------|--------------------------|--------------------------|
| `boolean` | `"true"` / `"false"` (strings) | `True` / `False` (Python bool) |
| `integer` | `str(attribute)` as-is | `int(str(attribute))` (cast to int) |
| `list` | `attribute.json()` (string) | `attribute.as_dict()` (dict/list) |
| `string` | `presentation % str(attribute)` | `str(attribute)` (no presentation) |
| unknown attrs | `str(attribute)` | `str(attribute)` |

---

## Missing `as_dict()` methods (5.0 patch is incomplete)

These attributes are `'list'` type in the representation dict. `_generate_dict` calls `attribute.as_dict()` on them, but no `as_dict()` was added:

- **`ExtendedCommunities`** (code 16) - will crash with `AttributeError`
- **`ExtendedCommunitiesIPv6`** (code 25) - will crash with `AttributeError`
- **`LargeCommunities`** (code 32) - will crash with `AttributeError`

Also missing: `LinkState` and `PrefixSid` sub-attributes that have `as_dict()` on the container but some leaf classes only have `json()`.

---

## BGP-LS NLRI classes

All follow the same pattern: `as_dict()` adds `"name"`, `"family"`, `"parsed": True` vs `json()` which uses `"ls-nlri-type"` key for the name.

Key naming difference: `json()` uses `"ls-nlri-type"` while `as_dict()` uses `"name"`.

---

## Summary: 5.0 before vs after

| Area | Change Type | Description |
|------|------------|-------------|
| rib.py top-level | Breaking | Array-of-routes under neighbor IP changed to flat per-route object with `peer` + `route` |
| INET key naming | Breaking | `"nlri"` renamed to `"prefix"` |
| Flow rules | Breaking | Flat rules moved under `"rules"` key |
| Flow nexthop key | Breaking | `"next-hop"` renamed to `"nexthop"` |
| Flow string field | Breaking | `"string"` compatibility field removed |
| EVPN `"raw"` field | Breaking | Present in json(), absent in as_dict() |
| EVPN null handling | Different | as_dict() always includes `"ip"`, `"nexthop"` (as null); json() omits when empty |
| EVPN `"maclen"` | Additive | New field in as_dict(), not in json() |
| nexthop everywhere | Additive | as_dict() includes nexthop on all NLRI types; json() does not |
| family everywhere | Additive | as_dict() includes structured family on all NLRI types; json() does not |
| BGP-LS name key | Breaking | `"ls-nlri-type"` changed to `"name"` |
| Label format | Different | Manual `[[value, raw?]]` vs `label.json()` |
| RD handling | Different | `self.rd._str()` or None vs `self.rd.json()` fragment |
| Attribute booleans | Different | Python `True/False` vs JSON `"true"/"false"` |
| Attribute integers | Different | Python `int` vs JSON string |
| ExtendedCommunities | Bug | Missing `as_dict()`: will crash at runtime |
| LargeCommunities | Bug | Missing `as_dict()`: will crash at runtime |
| `'multiple'` handler | Missing | `_generate_dict` has no `'multiple'` case |

---

## Where main additionally diverges

Main's codebase has structural differences from 5.0 beyond the patch:

1. **File layout**: main uses `collection.py` (wire/semantic separation) instead of `attributes.py`
2. **`'multiple'` type**: main's representation dict does not use `'multiple'` for any attribute (same as 5.0), but main's `_generate_json` has the handler
3. **ClusterList**: main stores packed bytes and unpacks lazily; 5.0 stores list of objects
4. **EVPN base json()**: main includes `route_code` property; 5.0 uses `self.CODE`
5. **No `as_dict()` on main**: main has zero `as_dict()` methods on NLRI/attribute classes (only on `NeighborTemplate`)
6. **rib.py**: main uses `async def callback()` with `asyncio.sleep(0)`; 5.0 uses generator `yield True/False`
7. **rib.py bugs**: main has the same pre-patch bugs as 5.0 (cidr.prefix() only for inet, pointless split, growing array)

thomas-mangin:
@brijohn I would prefer not to rollback the change, it is fine but we will have to keep backward compatibility for 5.0 users, they can not have the format change under their feet like that.

thomas-mangin:
Bug in the PR: missing as_dict() on 3 community types

  _generate_dict dispatches all 'list' representation types to attribute.as_dict(). Three community container classes are 'list' type but have no as_dict() method:
```
  ┌─────────────────────────┬──────┬──────────────────┐
  │        Attribute        │ Code │    as_dict()?    │
  ├─────────────────────────┼──────┼──────────────────┤
  │ Communities (initial)   │ 8    │ Yes (PR adds it) │
  ├─────────────────────────┼──────┼──────────────────┤
  │ ExtendedCommunities     │ 16   │ Missing          │
  ├─────────────────────────┼──────┼──────────────────┤
  │ LargeCommunities        │ 32   │ Missing          │
  ├─────────────────────────┼──────┼──────────────────┤
  │ ExtendedCommunitiesIPv6 │ 25   │ Missing          │
  ├─────────────────────────┼──────┼──────────────────┤
  │ ClusterList             │ 9    │ Yes (PR adds it) │
  ├─────────────────────────┼──────┼──────────────────┤
  │ ASPath                  │ 2    │ Yes (PR adds it) │
  ├─────────────────────────┼──────┼──────────────────┤
  │ LinkState               │ 29   │ Yes (PR adds it) │
  ├─────────────────────────┼──────┼──────────────────┤
  │ PrefixSID               │ 40   │ Yes (PR adds it) │
  └─────────────────────────┴──────┴──────────────────┘
```

On the exabgp PR: Add as_dict() to ExtendedCommunities, ExtendedCommunitiesIPv6, and LargeCommunities. The pattern matches Communities.as_dict():
```
  def as_dict(self):
      return [str(community) for community in self.communities]
```

thomas-mangin:
@brijohn adding info to the thread so I can use Claude or Codex to change the code later if required.

brijohn:
Well the missing as_dict in in LargeCommunities and ExtendedCommunities/IPv6 is not really bug in this PR, the as_dict is included in the basic Communities class from which the other three derive so it is not really missing they will just use the base class. We are using this with both Large and Extended communities and there is no crash.

I think the only real breaking change would be how inet routes are displayed prior to the patch vs after. Any other route type was already broken and casing the command to return an error.

Issue #1218, opened 2024-07-09
feature: Paths Limit for Multiple Paths in BGP


ton31337:
https://datatracker.ietf.org/doc/html/draft-abraitis-idr-addpath-paths-limit

E.g.: FRR already has support for this draft.

thomas-mangin:
*Posted by Claude (AI assistant) working on Thomas's machine, not written by Thomas.*

PATHS-LIMIT (draft-abraitis-idr-addpath-paths-limit-04) is implemented on `main` and will ship with 6.0. It is not in 5.0.

It is configured per family inside the `add-path` block, using the `limit` keyword:

```
neighbor 10.0.0.1 {
    router-id 10.0.0.2;
    local-address 10.0.0.2;
    local-as 65500;
    peer-as 65501;

    capability {
        add-path send/receive;
    }

    family {
        ipv4 unicast;
        ipv6 unicast;
    }

    add-path {
        ipv4 unicast limit 4;
        ipv6 unicast;
    }
}
```

Each line inside the `add-path` block takes one of two forms:

```
afi safi;              # ADD-PATH only, no paths-limit
afi safi limit N;      # ADD-PATH plus PATHS-LIMIT, N from 1 to 65535
```

The limit above says "send me at most 4 IPv4 unicast paths per prefix", and leaves IPv6 unicast with ADD-PATH and no limit.

Supported families are the ones ADD-PATH already covers: `ipv4/ipv6 unicast`, `ipv4/ipv6 labeled-unicast` (`nlri-mpls` also accepted) and `ipv4/ipv6 mpls-vpn`. A limit configured for any other family is parsed and then ignored, since neither capability is advertised for it.

On the sending side, a peer's advertised limit is enforced per prefix and holds across update batches, replacements and route refreshes. ExaBGP runs no best path selection, so the paths that get through are simply the ones announced first; the rest are held, and withdrawing an advertised path promotes one of them. Both events are logged under the `rib` category, with `exabgp_log_rib=true`:

```
rib.paths_limit.withheld family=... prefix=... limit=...
rib.paths_limit.promoted family=... prefix=... limit=...
```

On the receiving side, a peer that exceeds the limit we advertised gets a warning logged, once per prefix per family. The paths are not dropped and the session is not closed. It can be turned off with `exabgp_bgp_paths_limit_audit=false`.

Full documentation, including the wire format and the negotiation corner cases, is in `doc/user/paths-limit.md`.

PR #1387, opened 2026-04-27
Implement BGP paths limit capability

jamestiotio:
Implement the capability as specified by the `draft-abraitis-idr-addpath-paths-limit-04` document.

Corresponding unit tests are also added to validate the functionality of this feature.

### DIFF
--- a/etc/exabgp/conf-cap-paths-limit.conf
+++ b/etc/exabgp/conf-cap-paths-limit.conf
@@ -1,0 +1,16 @@
+neighbor 127.0.0.1 {
+	router-id 10.0.0.2;
+	local-address 127.0.0.1;
+	local-as 65533;
+	peer-as 65533;
+	hold-time 180;
+
+	capability {
+		add-path send/receive;
+		paths-limit 10;
+	}
+
+	family {
+		ipv4 unicast;
+	}
+}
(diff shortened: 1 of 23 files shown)


jamestiotio:
This closes #1218. First time contributing to ExaBGP, so apologies for any beginner mistakes.

thomas-mangin:
Thank you @jamestiotio
I will look into it in the next few days.

thomas-mangin:
The code was merged but then changed to match a different configuration format (and some issues were found by Claude which resolved them).

thomas-mangin:
Only for information, the findings were:

🔴 prefix_index() only implemented for INET

NLRI.prefix_index() falls back to self.index(), which includes the path-ID. For VPN, EVPN, FlowSpec, and other AddPath-capable families, paths-limit enforcement and audit will silently malfunction (each path-ID treated as a separate prefix, so the limit is never hit). Either implement prefix_index() on other NLRI subclasses or document this as an IPv4/IPv6 unicast-only feature.

🟡 Unbounded memory in _path_counts/_path_warned

IncomingRIB._path_counts grows with every unique prefix seen over the session lifetime. Withdrawn prefixes are cleaned up, but a peer that announces-then-withdraws millions of distinct prefixes will accumulate warned-set entries until clear(). Not critical since audit is optional, but worth noting.

🟡 unpack_capability loop repeats data = data[5:] three times

pathslimit.py:250,256,259 - could restructure to slice once at loop end. Minor readability nit.

🟡 Configuration output in neighbor.py:351-354

The paths-limit block emission mixes + concatenation with f-string implicit concatenation. Works, but fragile if someone later edits the surrounding f-string.

thomas-mangin:
Thank you for your help.

PR #1388, opened 2026-04-27
Remove md5 socket option when md5 password is cleared

brijohn:
This fixes an error under linux  where the listening socket will not disable md5 authentication.

This is a bit of an edge case, but if you start exabgp with an md5 password set and then remove the md5 option and reload the config, the md5 socket option is left on the listening socket. This will cause a BGP peering session to fail if exabgp is operating as a passive peer.

### DIFF
--- a/src/exabgp/reactor/listener.py
+++ b/src/exabgp/reactor/listener.py
@@ -63,8 +63,7 @@ class Listener:
                 continue
             if local_port != port:
                 continue
-            if use_md5:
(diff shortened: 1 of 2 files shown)


thomas-mangin:
Thank you merged

thomas-mangin:
Hi @brijohn - well spotted - thank you, it is in 5.0 abd forward ported to main too.

PR #1390, opened 2026-05-05
Add api show-routes-format env to allow using old behaviour

brijohn:
This updates my previous PR #1385 to allow selecting which json version "show adj-rib" gives.  If the environment api.show-routes-format is set to v1 you will get the original output, but still only for inet based routes. When set to v2 it will use the format supporting all route types. V1 mode has been at least fixed so it does not actually generate exceptions when non inet NLRI is encountered and instead it just skips it. 

This also patches another issue I discovered as well. I was using pattern 'if self.rd is not RouteDistinguisher.NORD' in my code. However there are places in the code that perform a deepcopy of a change object which in turn makes a copy of the rd object and in doing so destroys the IS relationship in python since the rd is no longer the same object as NORD. To fix this i implemented the `__copy__` and `__deepcopy__` functions in the RouteDistinguisher class so if an rd equals b'' then it returns the NORD singleton.

### DIFF
--- a/src/exabgp/bgp/message/update/nlri/qualifier/rd.py
+++ b/src/exabgp/bgp/message/update/nlri/qualifier/rd.py
@@ -54,6 +54,16 @@ class RouteDistinguisher:
     def __len__(self):
         return self._len
 
+    def __copy__(self):
+        if self == RouteDistinguisher.NORD:
+            return RouteDistinguisher.NORD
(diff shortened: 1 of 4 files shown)


thomas-mangin:
I am not ignoring the PR, I am currently in Japan travelling and will look into it when I am back next week

PR #1395, opened 2026-05-22
Install and use sysusers.d config file

bluca:
sysusers.d config files allow a package to use declarative configuration instead of manually written maintainer scripts. This also allows image-based systems to be created with /usr/ only, and also allows for factory resetting a system and recreating /etc/ on boot. https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -4,6 +4,7 @@ Section: net
 Priority: optional
 Build-Depends: debhelper-compat (= 13),
                dh-python,
+               dh-sequence-installsysusers,
                python3-all,
                python3-setuptools,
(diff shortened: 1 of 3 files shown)


thomas-mangin:
@vincentbernat should I remove the debian folder on main as you manage it on the branches and I can not review the patch of @bluca as it would affect your work.

vincentbernat:
Yes, you can remove it. But @bluca patch is already on the right branch.

thomas-mangin:
Thank you @vincentbernat for your help - much appreciated.

PR #1401, opened 2026-06-25
Add DDoS auto-mitigation example with FlowSpec

jacob-masse:
Adds a practical example showing how to use ExaBGP for automated DDoS mitigation via FlowSpec. The example receives attack alerts via webhook and dynamically announces/withdraws FlowSpec rules. Demonstrates integration with Flowtriq but the pattern works with any detection system.

## What's included

- `etc/exabgp/api-ddos-flowspec.conf` -- ExaBGP configuration with FlowSpec family and process setup
- `etc/exabgp/run/api-ddos-flowspec.run` -- Python process script: HTTP webhook listener that translates attack alerts into FlowSpec announce/withdraw commands
- `etc/exabgp/run/README-ddos-flowspec.md` -- Setup guide with architecture diagram, curl examples, and payload reference

## How it works

A lightweight HTTP server (stdlib only, no dependencies) runs as an ExaBGP process. When a DDoS detection system sends a POST with attack details (source/destination prefix, protocol, port), the script writes the corresponding FlowSpec command to stdout for ExaBGP to announce upstream. On attack clear, it withdraws the route.

The webhook interface is generic -- works with any system that can POST JSON (Flowtriq, custom sFlow analyzers, Zabbix, etc.).

### DIFF
--- a/doc/ddos-flowspec.md
+++ b/doc/ddos-flowspec.md
@@ -1,0 +1,311 @@
+# DDoS Mitigation with FlowSpec
+
+This guide explains how to use ExaBGP as an automated FlowSpec controller
+for DDoS mitigation.  A webhook listener receives attack alerts from a
+detection system and translates them into BGP FlowSpec rules that upstream
+routers enforce at the network edge.
+
+## What is FlowSpec?
+
+BGP FlowSpec (RFC 5575, updated by RFC 8955) extends BGP to distribute
+traffic filtering rules alongside routing information.  Instead of
+blackholing an entire prefix, FlowSpec lets you describe attack traffic
+precisely -- by source, destination, protocol, port, packet length, and
+more -- and instruct routers to drop, rate-limit, or redirect only the
+matching packets.
+
+This makes FlowSpec ideal for surgical DDoS mitigation: you keep
+legitimate traffic flowing while the attack is filtered at the router
+level, before it reaches your servers.
+
+## Prerequisites
+
+1. **Router support** -- Your upstream or edge routers must support BGP
+   FlowSpec.  Most modern Juniper (MX, PTX), Cisco (IOS-XR), Nokia (SR OS),
+   and Arista (EOS) platforms do.  Check your vendor documentation for the
+   required license or feature set.
+
+2. **BGP session** -- ExaBGP must have an iBGP or eBGP session to the
+   router(s) where you want FlowSpec rules installed.  The session needs
+   `ipv4 flow` (and optionally `ipv6 flow`) address families negotiated.
+
+3. **Validation mode** -- Some routers require the FlowSpec originator to
+   also be the best-path next-hop for the destination prefix, or they will
+   reject the rule.  Check your router's FlowSpec validation settings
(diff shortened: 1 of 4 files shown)


thomas-mangin:
Thank you for this, @jacob-masse

It is a good example, well put together. As it relates to your product, I have noticed over the years that people do not see examples as examples. They copy them into production almost verbatim, defaults and all, so anything shipped should be solid enough to survive that. 

There are two points I would like tightened up:

1. Validate the input before it reaches the API. Right now source, destination, protocol, and port go straight from the JSON body into the command string.
2. Make the safe choice the default. The listener binds to 0.0.0.0 with no authentication. The README mentions tightening this for production, but since this is what people will copy, I would rather the shipped default be the safe one: bind to 127.0.0.1 and check a simple API key, with a note explaining how to open it up if they really need to.

A few smaller things while you are in there, none of them blocking on their own:

 - The time.sleep(0.5) guesses how long ExaBGP needs to start up. Since you are already reading ExaBGP's messages on stdin, it would be safer to start the listener when ExaBGP reports the session is up, rather than after a fixed delay.
- ThreadingHTTPServer instead of HTTPServer would stop alerts from serializing, which matters precisely when you are under attack.
- The DNS amplification example matches destination-port =53, but reflected replies come from source port 53, so source-port is likely what you want there. Worth getting right since people will model their own rules on it.

If you can fold those in, I am glad to merge it. Thanks again for taking the time to write it up properly, the documentation in particular is the part people usually skip.

thomas-mangin:
also you may want to have a look at my latest project (WIP - no release yet)
https://github.com/ze-software/ze/

jacob-masse:
Thanks for the thorough review, Thomas. Pushed an update addressing everything:

- **Input validation**: source/destination validated as CIDR via `ipaddress`, protocol checked against known set, port range 1-65535
- **Safe defaults**: binds 127.0.0.1, optional API key auth via `API_KEY` env var, documented how to open it up
- **Session-up wait**: replaced `time.sleep(0.5)` with a threading.Event that fires on ExaBGP's connected/up message on stdin
- **ThreadingHTTPServer**: switched from HTTPServer so alerts don't serialize under load
- **DNS amplification fix**: added `source_port` field, fixed the example to use source-port (reflected replies)
- **Documentation**: added `doc/ddos-flowspec.md` with full setup guide, common attack patterns, and troubleshooting

Let me know if anything else needs adjusting.

thomas-mangin:
No worries, most of the work was @claude 😉

thomas-mangin:
Nice product you are building.

I got Claude to update https://github.com/Exa-Networks/exabgp/wiki/DDoS-Mitigation but it could have made a mistake about how things works - if so let me know please.

thomas-mangin:
@jacob-masse can you please let me know if https://github.com/Exa-Networks/exabgp/wiki/DDoS-Mitigation is correct about your product please :-)

jacob-masse:
Hi Thomas,

Thank you! I appreciate the kind words and the wiki addition.

Your entry is accurate and looks great to me!

PR #1410, opened 2026-08-11
Dockerfile: pin the uv builder image, and actually drop the apt lists

sujeito-operator:
Two small build-hygiene fixes, one commit each, so you can take either on its own or
neither.

### 1. `COPY --from=ghcr.io/astral-sh/uv:latest` -> `:0.12.3`

`COPY --from=` on line 16 names an external image rather than an earlier stage, so
`:latest` leaves the tool that builds the wheel floating: a build today and a build next
month can use different uv versions, and nothing in the tree records which one produced a
given image.

That's the same class of thing you already pin everywhere else — dependabot keeps
`actions/checkout`, `actions/setup-python` and `github/codeql-action` on exact versions
for you. This is the one build input it can't see, because dependabot doesn't read
`COPY --from`. So pinning it does mean carrying the bump by hand. If you'd rather keep it
floating for that reason, just drop this commit — it's your build and I have no stake in
the answer.

### 2. `apt-get clean` doesn't remove the package indices

`apt-get clean` empties `/var/cache/apt/archives` — the downloaded `.deb` files. The
package *indices* that `apt-get update` fetched live in `/var/lib/apt/lists` and survive
it, so they ship in the final image. Adding `rm -rf /var/lib/apt/lists/*` to the same
`RUN` finishes what the `clean` is already there to do.

On the size, and I'd rather show the working than assert a number: **I don't have a
Docker daemon on this machine, so this is not a `docker images` diff.** It's derived from
the index sizes Debian publishes in its own `Release` file for `bookworm
main/binary-amd64`:

| form | size |
|---|---|
| `Packages` (unpacked) | 47.74 MiB |
| `Packages.gz` | 11.52 MiB |
| `Packages.xz` | 8.38 MiB |

The official Debian images set `Acquire::GzipIndexes "true"` (debuerreotype does this
deliberately, precisely because Dockerfiles tend not to delete the lists), so apt keeps
the compressed form rather than unpacking to 47.74 MiB. **The saving is therefore roughly
9–12 MB, not the ~48 MB you'd get if you quoted the unpacked figure.** Modest, but it's
on every pull, and the fix is one clause.

Neither commit changes what's installed: `iproute2` and `dumb-init` are installed before
the `rm`, and the `# syntax=docker/dockerfile:1.4` frontend is untouched.

---

Both were found by a Dockerfile linter I maintain, run over public repositories — the
`COPY --from` case was a rule it was *missing*, which is how this ended up worth a patch
rather than a note. The patch itself was prepared by an automated agent and re-checked
against `main` by hand before opening; saying so up front seems better than not. Happy to
close it if this isn't useful to you.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 
 FROM python:3.13-bookworm AS builder
 
-COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+COPY --from=ghcr.io/astral-sh/uv:0.12.3 /uv /uvx /bin/
 
 WORKDIR /build
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,8 @@ ENV PATH=/opt/exabgp/bin:$PATH
 # install deps
 RUN apt-get update \
     && apt-get install -y iproute2 dumb-init \
-    && apt-get clean
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
 
 # Add ExaBGP
 RUN useradd -r exa \
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 
 FROM python:3.13-bookworm AS builder
 
-COPY --from=ghcr.io/astral-sh/uv:0.12.3 /uv /uvx /bin/
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 
 WORKDIR /build
 


thomas-mangin:
I agree the rm would help but not the pinning of uv. I do not track uv security or release and uv as a tool should not change its behaviour so latest version would have no difference.

If you update the patch to only include the remove I will merge it, if not I will likely add the line myself and commit it and then close this PR.

sujeito-operator:
Done — the patch is now just the `rm`.

Your reasoning holds: `uv` is a build-time tool whose behaviour shouldn't shift between releases, you don't track its advisories, and a pin you don't watch is a bump someone has to remember. I'd rather drop it than leave you maintaining a version number for a guarantee you weren't asking for.

What's left is one line — `rm -rf /var/lib/apt/lists/*` in the same `RUN` as the `apt-get install`, so the lists don't ship inside the image.

Thanks for taking the time to say which half you wanted rather than just closing it.

thomas-mangin:
Always nice to work with a well behaved AI

PR #1393, opened 2026-05-15
feat: Add SR-Policy (SAFI 73) support

manoharan-nexthop:
Addresses https://github.com/Exa-Networks/exabgp/issues/1204

**SR-Policy NLRI (SAFI 73)**:
- Full encode/decode for IPv4 and IPv6 SR-Policy NLRI
- Distinguisher, color, endpoint fields per RFC 9256
- Family negotiation via 'family { ipv4 sr-policy; ipv6 sr-policy; }'
- SAFI.sr_policy = 73 registered in protocol/family.py

**SR-Policy Tunnel Encapsulation Path Attribute (RFC 9830, RFC 9831)**:
- SR Preference, Priority, Binding SID (MPLS), SRv6 Binding SID
- SR Policy Name, SR Candidate Path Name
- SR Segment List with all segment types per RFC 9831: Type A  (1)  MPLS label Type B  (13) SRv6 SID [+ Endpoint Behavior] Type C  (3)  IPv4 Node + SR Algorithm [+ MPLS SID] Type D  (4)  IPv6 Node + SR Algorithm [+ MPLS SID] Type E  (5)  IPv4 Node + Local Interface ID [+ MPLS SID] Type F  (6)  IPv4 Adjacency [+ MPLS SID] Type G  (7)  IPv6 Link-local Adjacency + Interface IDs [+ MPLS SID] Type H  (8)  IPv6 Adjacency [+ MPLS SID] Type I  (14) IPv6 Node + SR Algorithm [+ SRv6 SID [+ EB]] Type J  (15) IPv6 Link-local Adjacency + Interface IDs + Algorithm [+ SRv6 SID [+ EB]] Type K  (16) IPv6 Adjacency + SR Algorithm [+ SRv6 SID [+ EB]]

**Configuration**:
- 'announce ipv4/ipv6 sr-policy' inline syntax via ParseAnnounce.register_family
- 'static { sr-policy ... }' inline syntax via ParseStatic.register_command
- conf-sr-policy.conf: functional test config (127.0.0.1, AS 65533)
- conf-sr-policy-simple.conf: minimal working example
- conf-sr-policy-static.conf: static section inline examples
- conf-sr-policy-all-types.conf: all segment types A-K in one file

**Fixes included**:
- configuration/core/parser.py: fix backslash line continuation (strip trailing \ before concatenating continued lines)
- reactor/api/command/rib.py: fix to_json for non-prefix NLRIs (use str(nlri) fallback when .cidr attribute absent)
[...]

PR #1417, opened 2026-08-26
fix: conform SR Policy sub-TLV bits to RFC 9830/9831, add ENLP and V-Flag

manoharan-nexthop:
Follow-up to #1393, addressing the S-bit review comment and a full flag/type audit against RFC 9830 and RFC 9831.

## RFC conformance fixes

- **Segment MPLS label S bit**: RFC 9830 §2.4.4.2.1 says the S bit "MUST be zero upon transmission and MUST be ignored upon reception". The code set S=1 on the last MPLS segment as if it were an RFC 3032 bottom-of-stack marker — but this is control-plane signaling; the headend builds the actual label stack. Now always transmitted as zero (types A, C–H); decoded values still surface in JSON.
- **Binding SID label entry**: TC/S/TTL are RESERVED per RFC 9830 §2.4.2 — now zero (S=1 was hardcoded). Also removed a transmitted `0x10` flag that does not exist in the "SR Policy Binding SID Flags" registry (only S-Flag 0x80 and I-Flag 0x40 are assigned; unassigned bits must be zero). BSID presence is signaled by length.
- **SID Specified S-Flag (0x20, RFC 9831)**: now set in segment types C–K whenever the optional SID field is present; it was never transmitted.

## New support

- **ENLP sub-TLV** (type 14, RFC 9830 §2.4.5): `enlp <push-ipv4|push-ipv6|push-ipv4-ipv6|no-push>` (or numeric 1–4).
- **V-Flag** (SID verification, 0x80): per-segment `verification` keyword, valid on all segment types; emitted in JSON as `"verification": true` so decode→re-encode round-trips.

## Testing

- `./qa/bin/test_everything`: all 23 suites pass
- New unit tests: S=0 on transmission (incl. decode-with-S=1 repacks as S=0), BSID flags byte, S-Flag set/unset with/without SID, ENLP exact bytes + round-trip, V-Flag pack + JSON
- Functional encoding test extended with an announce exercising `enlp` + `verification`


thomas-mangin:
Thank you for your help.

manoharan-nexthop:
> Thank you for your help.

I see the PR is closed, do we not require this change

thomas-mangin:
I merged using my local machine and not github
https://github.com/Exa-Networks/exabgp/commit/6e164b1453b4c35ece9f4d4106fea6e591e2ca3d
And continued on the work
https://github.com/Exa-Networks/exabgp/commit/3571bc849d77caa1358d5a7dc0fcbf15c137cc57

manoharan-nexthop:
> I merged using my local machine and not github [6e164b1](https://github.com/Exa-Networks/exabgp/commit/6e164b1453b4c35ece9f4d4106fea6e591e2ca3d) And continued on the work [3571bc8](https://github.com/Exa-Networks/exabgp/commit/3571bc849d77caa1358d5a7dc0fcbf15c137cc57)

great, and thank you again for catching this!

PR #1419, opened 2026-08-26
Add the ability to use Link Local Ipv6 addresses which are scoped to an interface as BGP neighbors and local addresses

anish-n:
Why I made this change?
Failure to bringup a Link Local IPv6 neighbor by following the steps listed in https://github.com/exa-networks/exabgp/wiki/Neighbor-Configuration#link-local-next-hop-600, with neighbor configuration containing the following:

```
neighbor fe80::1%eth0 {
    router-id 192.168.1.2;
    local-address fe80::2%eth0;
    local-as 65001;
    peer-as 65000;

    capability {
        link-local-nexthop enable;
        link-local-prefer true;
    }
}
```

Changes made:
1. Modify the config parser to accept zones
2. Modify sock.connect call to use the scope_id from ```neighbor <ip>%<intf>```
3. Modify sock.bind call to use the scope_id from ```local-address <ip>%<intf>```

Potential documentation update needed: Supporting zone field on local address ```local-address fe80::2%eth0``` after this PR merges

### DIFF
--- a/src/exabgp/protocol/ip/__init__.py
+++ b/src/exabgp/protocol/ip/__init__.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from exabgp.bgp.message.open.capability.negotiated import Negotiated
 
 
+
 class IPFactory(Protocol):
     """Protocol for IP subclass constructors that accept packed data.
 
@@ -71,6 +72,7 @@ class IP(IPBase):
     def init(self, packed: Buffer) -> IP:
         self._packed = packed
         self.afi = IP.toafi(IP.ntop(packed))
+        self._zone = ''
         return self
 
     def __iter__(self) -> Iterator[str]:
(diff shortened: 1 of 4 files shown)


thomas-mangin:
Thank you I will review in a bit

thomas-mangin:
Thank you for this. You were right, the wiki page does not work. That said, very sorry, I am not going to merge the patch.

A zone has no wire representation, so carrying one there changes `top()` and it makes no sense.

`source-interface` was already applied to the outgoing socket through SO_BINDTODEVICE, and on Linux. The kernel wants `sk_bound_dev_if` set before a link-local address can be bound or connected, SO_BINDTODEVICE sets it.

```
neighbor fe80::1 {
    router-id 192.168.1.2;
    local-address fe80::2;
    source-interface eth0;
    local-as 65001;
    peer-as 65000;

    capability {
        link-local-nexthop enable;
    }
}
```

Your report is what led me to find that the keyword was only half wired. `SO_BINDTODEVICE` unfortunately is a Linux only feature, so none of this works on BSD or macOS.

The wiki page you followed has been rewritten as it was simply wrong. Sorry for the time that cost you.

Thank you again for the report and the patch. Hopefully `source-interface` will do what you need.

https://github.com/Exa-Networks/exabgp/commit/79031c57d7d71a09fc0f7df8c189455ddd002c25

anish-n:
> Thank you for this. You were right, the wiki page does not work. That said, very sorry, I am not going to merge the patch.
> 
> A zone has no wire representation, so carrying one there changes `top()` and it makes no sense.
> 
> `source-interface` was already applied to the outgoing socket through SO_BINDTODEVICE, and on Linux. The kernel wants `sk_bound_dev_if` set before a link-local address can be bound or connected, SO_BINDTODEVICE sets it.
> 
> ```
> neighbor fe80::1 {
>     router-id 192.168.1.2;
>     local-address fe80::2;
>     source-interface eth0;
>     local-as 65001;
>     peer-as 65000;
> 
>     capability {
>         link-local-nexthop enable;
>     }
> }
> ```
> 
> Your report is what led me to find that the keyword was only half wired. `SO_BINDTODEVICE` unfortunately is a Linux only feature, so none of this works on BSD or macOS.
> 
> The wiki page you followed has been rewritten as it was simply wrong. Sorry for the time that cost you.
> 
> Thank you again for the report and the patch. Hopefully `source-interface` will do what you need.
> 
> [79031c5](https://github.com/Exa-Networks/exabgp/commit/79031c57d7d71a09fc0f7df8c189455ddd002c25)

Ok thank you, understood, I will close the PR in that case.

PR #1422, opened 2026-09-06
fix: use os.read() to avoid BufferedReader/poll() data stranding

### LINKED ISSUE #1421: BufferedReader + poll() data stranding in processes.py (5.0 / 4.2)
## Summary

`proc.stdout.read(16384)` in the reactor's process reader goes through Python's `BufferedReader`, whose internal two-phase read path can consume more bytes from the kernel pipe than it returns. The excess sits in `BufferedReader`'s user-space buffer, invisible to `select.poll(0)`, which watches the kernel fd, causing route commands to strand silently.

**Affected code:**
- **5.0**: [`src/exabgp/reactor/api/processes.py:238–258`](https://github.com/Exa-Networks/exabgp/blob/82fc1a5535292a04d822ea4c47f3971fc10729a4/src/exabgp/reactor/api/processes.py#L238-L258)
- **4.2**: [`lib/exabgp/reactor/api/processes.py:216–235`](https://github.com/Exa-Networks/exabgp/blob/ee02a9e87d3e0a10d33231bf9b67bc4c3a0e496b/lib/exabgp/reactor/api/processes.py#L216-L235)
- **main**: the sync `received()` method still contains the pattern but is dead code since the asyncio migration made `received_async()` the only active reader.

**Affected versions:** 4.0.4 through 5.0.13 (latest release).

## Buggy Code (5.0 branch)

```python
# src/exabgp/reactor/api/processes.py L238-L258
poller = select.poll()
poller.register(
    proc.stdout,
    select.POLLIN | select.POLLPRI | select.POLLHUP | select.POLLNVAL | select.POLLERR,
)

ready = False
for _, event in poller.poll(0):
    if event & select.POLLIN or event & select.POLLPRI:
        ready = True
    elif event & select.POLLHUP or event & select.POLLERR or event & select.POLLNVAL:
        self._handle_problem(process)

if not ready:
    continue                                          # ← skips if poll says no data

buf = str(proc.stdout.read(16384), 'ascii')           # ← BufferedReader, not os.read
```

`poll()` and `BufferedReader.read()` talk about **different buffers**:

```
+---------------------------------------------------------------+
|  USER SPACE                                                   |
|  +----------------------------+                               |
|  | Python BufferedReader      |                               |
|  | [internal buffer (8 KB)] <-+-- proc.stdout.read(16384)     |
|  |         ^                  |  reads from HERE.             |
|  |         | fill-buffer      |  May hold data that           |
|  +---------|------------------+  poll() cannot see.            |
|            | raw.read()                                       |
+------------|--------------------------------------------------+
|  KERNEL SPACE                                                 |
|            v                                                  |
|  +----------------------------+                               |
|  | Pipe buffer (64 KB)        |  select.poll(0) watches HERE. |
|  +----------------------------+                               |
+---------------------------------------------------------------+
```

## Root Cause

`subprocess.Popen` with `stdout=subprocess.PIPE` wraps the pipe fd in a `BufferedReader` (buffer_size=8192). The fd is set to `O_NONBLOCK`.

`BufferedReader.read(16384)` issues a raw read from the kernel. If the first raw read returns fewer than 16384 bytes but more than 8192 (a partial read), it enters a [two-phase path in CPython](https://github.com/python/cpython/blob/main/Lib/_pyio.py#L1132-L1147):

1. **Phase 1**: Direct read: `raw.read()` → N bytes (8192 < N < 16384)
2. **Phase 2**: Fill buffer: since `(16384 - N) < 8192`, it calls `raw.read(8192)` → M bytes, takes `(16384 - N)` from M, keeps `(M - 16384 + N)` hidden in the internal buffer.

On the next reactor cycle, `poll(0)` sees an empty kernel fd and returns "not ready", skipping the process, even though `BufferedReader` has data waiting. Mid-batch this is harmless (new writes keep triggering poll), but **at the batch tail** , once the writer has flushed all commands, no further writes will re-trigger poll, and the stranded commands are lost. This makes the bug **flaky**: it only triggers when kernel scheduling aligns the writer's final flush with `BufferedReader`'s fill-buffer window.

## How It Was Introduced

| Date | Commit | Change |
| :--- | :--- | :--- |
| 2017-07-20 | [`898d54ab`](https://github.com/Exa-Networks/exabgp/commit/898d54ab) | `readline()` → `read()` (unbounded) : safe |
| **2017-07-20** | [**`51aeeaf2`**](https://github.com/Exa-Networks/exabgp/commit/51aeeaf2) | **`read()` → `read(16384)` : bug introduced** |
| 2019-12-10 | [`6401704d`](https://github.com/Exa-Networks/exabgp/commit/6401704d) | `select.select` → `select.poll` : same bug, different poller |

The unbounded `read()` drained all kernel data, so nothing could strand. Adding the 16384 cap created the gap between poll's view and `BufferedReader`'s consumption.

## Fix

```diff
-  buf = str(proc.stdout.read(16384), 'ascii')
+  buf = str(os.read(proc.stdout.fileno(), 16384), 'ascii')
```

`os.read(fd, 16384)` bypasses `BufferedReader`, reads directly from the kernel fd, never consumes more than it returns.

## Environment

- Observed in ExaBGP 4.2.x and 5.0.x on Linux
- Python 3.8+
- Reproducible when a helper process writes large batches of route commands via stdout pipe

rajkumar1-arista:
Fixes #1421

## Summary

- Replace `proc.stdout.read(16384)` with `os.read(proc.stdout.fileno(), 16384)` in the sync process reader to prevent `BufferedReader`'s internal buffer from hiding data from `select.poll()`
- Fixes silent route command stranding at the tail of large batches

## Root Cause

`proc.stdout.read(16384)` goes through Python's `BufferedReader` (buffer_size=8192). When a partial read returns between 8193–16383 bytes, `BufferedReader` enters a two-phase path that fills its internal buffer with bytes beyond what it returns to the caller. On the next reactor cycle, `poll(0)` sees an empty kernel fd and skips the process — those bytes are stranded until new data arrives on the pipe.

This is harmless mid-batch (new writes keep triggering poll), but **fatal at the batch tail**: once the writer has flushed all commands, no further writes will re-trigger poll, and the stranded commands are lost.

`os.read(fd, 16384)` bypasses `BufferedReader` entirely — it reads directly from the kernel fd, never consuming more than it returns, so `poll()` and `read()` always agree.

## The bug was introduced in

| Date | Commit | Change |
| :--- | :--- | :--- |
| 2017-07-20 | `898d54ab` | `readline()` → `read()` (unbounded) — safe |
| **2017-07-20** | **`51aeeaf2`** | **`read()` → `read(16384)`** — bug introduced |

The unbounded `read()` was safe because it drains everything from the kernel. Adding the 16384 cap created the mismatch.

### DIFF
--- a/src/exabgp/reactor/api/processes.py
+++ b/src/exabgp/reactor/api/processes.py
@@ -255,7 +255,7 @@ class Processes:
                     # Calling next() on Linux and OSX works perfectly well
                     # but not on OpenBSD where it always raise StopIteration
                     # and only read() works (not even readline)
-                    buf = str(proc.stdout.read(16384), 'ascii')
+                    buf = str(os.read(proc.stdout.fileno(), 16384), 'ascii')
                     if buf == '' and poll is not None:
                         # if proc.poll() is None then
                         # process is fine, we received an empty line because


thomas-mangin:
*Posted by Claude (AI assistant) working on Thomas's machine, not written by Thomas.*

Review of the one line change in `Processes.received()`:

```diff
-  buf = str(proc.stdout.read(16384), 'ascii')
+  buf = str(os.read(proc.stdout.fileno(), 16384), 'ascii')
```

## The change is correct

I reproduced the root cause rather than taking #1421 on trust. With an 8192 byte `BufferedReader` on an `O_NONBLOCK` pipe holding 17000 bytes, `read(16384)` returns 16384 bytes and leaves 616 in the user space buffer; `select.poll(0)` on the fd then returns `[]` while `os.read` on the same fd gives `EAGAIN`. That is exactly the stranding path: `received()` sees `not ready`, does `continue`, and those bytes are only released if the child writes again, which never happens at the tail of a batch.

Checks that pass:

- `os` is already imported (line 10), and the fd is explicitly set `O_NONBLOCK` in `_start` (line 179), so `os.read` cannot block.
- Line 258 is the only reader of `proc.stdout` in `src/`, so nothing else depended on the `BufferedReader` and no data can be stranded in it.
- `BlockingIOError` / EAGAIN is now reachable where the old code returned `None`. It lands in `except OSError`, and `error.block` in `src/exabgp/reactor/network/error.py` covers `EAGAIN`, `EWOULDBLOCK`, `EINTR`, `EDEADLK`, so it is a benign `pass`. This is actually an improvement: the old `read()` returning `None` made `str(None, 'ascii')` raise `TypeError`, which neither the inner `except OSError` nor the outer `except (CalledProcessError, OSError, ValueError)` catches, so it would take the reactor down.
- EOF is unchanged: `os.read` returns `b''` exactly as the buffered read did, so the `buf == '' and poll is not None` branch still works.
- Short reads and a command split across a read boundary are handled by the existing `self._buffer[process]` carry over, and ASCII is single byte so there is no new split decode hazard.
- The single caller (`src/exabgp/reactor/loop.py:433`) drains the generator to exhaustion with no `break`, so `self._buffer[process] = raw` always runs and the partial line survives.

## Requested change: a regression test

The repository rule is explicit that a bug fix comes with the test that fails without it, and this failure mode is silent data loss that a future cleanup could reintroduce by "simplifying" `os.read(proc.stdout.fileno(), n)` back to `proc.stdout.read(n)`. The functional suite will not catch a reintroduction, because the bug only fires when the writer's final flush lands in the fill buffer window.

A deterministic unit test is cheap: create a pipe, set `O_NONBLOCK`, wrap the read end in `io.BufferedReader(io.FileIO(fd), buffer_size=8192)`, write 17000 bytes, read 16384, then assert `poll(0)` reports nothing while bytes remain unread. `tests/unit/` on this branch is the right home, next to `test_process_state.py` and `test_text_api_stream.py`.

## Minor points

**The comments now document the wrong thing** (`processes.py:255-263`). Lines 255 to 257 still justify the call as "next() raises StopIteration on OpenBSD, only read() works", and line 262 still says "we're doing .read() on a non-blocking pipe". Neither records why the code must use the raw fd, which is precisely the knowledge a future refactor needs in order not to undo this. One line saying `os.read` is deliberate because `BufferedReader` can hold bytes that `poll()` on the fd cannot see would be enough.

**The fix is not propagated.** The 4.2 tree (`lib/exabgp/reactor/api/processes.py`) has the identical pattern and is untouched here. On `main` the same line survives in the sync `received()`; if that path is dead since the asyncio migration it would be better deleted than left as a live looking copy of a known bad pattern, otherwise a 5.0 to main merge quietly loses this fix.

**Optional alternative.** Passing `bufsize=0` to `subprocess.Popen` would make `proc.stdout` a raw `FileIO` and fix the same problem without the `fileno()` round trip. `os.read` is fine and matches what `main` already does, so this is not worth blocking on.

## One note for the maintainers, outside this PR

`main`'s asyncio reader is not affected by this bug: `_async_reader_callback` already uses `os.read(fd, 16384)` and `loop.add_reader` is level triggered on the same kernel fd, so `poll` and `read` agree there. Separately though, that handler's `except OSError` removes the reader before classifying the errno, so a transient EAGAIN would permanently stop reading from that process. That is a pre-existing `main` issue, not something this PR introduces.

https://claude.ai/code/session_01FvWnSxwXsRe6YncfSqRFk2

PR #1424, opened 2026-09-07
fix: backport BufferedReader/poll() data stranding fix to 4.2

### LINKED ISSUE #1421: BufferedReader + poll() data stranding in processes.py (5.0 / 4.2)

rajkumar1-arista:
## Summary

Cherry-pick of #1422 and its follow-up test commit to the 4.2 branch.

- Replace `str_ascii(proc.stdout.read(16384))` with `str_ascii(os.read(proc.stdout.fileno(), 16384))` to prevent `BufferedReader`'s internal buffer from hiding data from `select.poll()`
- Add test proving the raw-fd read is required and that `received()` returns every command

Cherry-picked from:
- [df00f00fd](https://github.com/Exa-Networks/exabgp/commit/df00f00fd0290edb37fb90b2084914910ec92dc8) — the fix
- [446669464](https://github.com/Exa-Networks/exabgp/commit/4466694642f7078445a44431de75e5e000509852) — the test + explanatory comment

Conflict resolution: 4.2 uses `str_ascii()` wrapper (vs `str(..., 'ascii')` on 5.0) and `exabgp.configuration.setup` for environment initialization (vs `exabgp.environment` on 5.0). Both adapted accordingly.

Fixes #1421

### DIFF
--- a/lib/exabgp/reactor/api/processes.py
+++ b/lib/exabgp/reactor/api/processes.py
@@ -232,11 +232,16 @@ class Processes(object):
                     # Calling next() on Linux and OSX works perfectly well
                     # but not on OpenBSD where it always raise StopIteration
                     # and only read() works (not even readline)
+                    # os.read() on the raw fd is deliberate: proc.stdout is a
+                    # BufferedReader which can consume more from the pipe than it
+                    # returns, and the leftover bytes sit in a user-space buffer
+                    # that the poll() above (which watches the kernel fd) can not
+                    # see, stranding commands at the tail of a batch.
                     buf = str_ascii(os.read(proc.stdout.fileno(), 16384))
                     if buf == '' and poll is not None:
                         # if proc.poll() is None then
                         # process is fine, we received an empty line because
-                        # we're doing .read() on a non-blocking pipe and
+                        # we're reading a non-blocking pipe and
                         # the process maybe has nothing to send yet
                         self._handle_problem(process)
                         continue
--- a/lib/exabgp/reactor/api/processes.py
+++ b/lib/exabgp/reactor/api/processes.py
@@ -232,7 +232,7 @@ class Processes(object):
                     # Calling next() on Linux and OSX works perfectly well
                     # but not on OpenBSD where it always raise StopIteration
                     # and only read() works (not even readline)
-                    buf = str_ascii(proc.stdout.read(16384))
+                    buf = str_ascii(os.read(proc.stdout.fileno(), 16384))
                     if buf == '' and poll is not None:
                         # if proc.poll() is None then
                         # process is fine, we received an empty line because
--- a/qa/tests/test_process_read_stranding.py
+++ b/qa/tests/test_process_read_stranding.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python3
+# encoding: utf-8
+
+"""The process reader must read the kernel fd, never the BufferedReader over it
+
+`Processes.received()` decides whether a helper process has anything to say by
+calling `select.poll(0)` on the process' stdout file descriptor. That asks the
+kernel about the pipe. If the read which follows goes through the `BufferedReader`
+which `subprocess.Popen` wraps the pipe in, the two disagree: `BufferedReader.read(n)`
+can drain more from the pipe than it hands back, and the excess sits in a user-space
+buffer that `poll()` can not see.
+
+Mid-batch that is harmless, the writer's next flush makes poll ready again and the
+hidden bytes come out. At the tail of a batch nothing follows, so the last commands
+are stranded and silently never executed. See issue #1421: the bug arrived in 2017
+when an unbounded `read()` (which could not strand anything) became `read(16384)`.
+
+`os.read(fd, n)` never consumes more than it returns, so poll and the read agree.
+"""
+
+import io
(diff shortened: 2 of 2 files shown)


thomas-mangin:
Thnk you for this work and sorry for refusing to accept it.
4.2 is not supported anymore, so I will close the PR. User should update to 5.0

